PR #142, opened 2016-10-17
fix bug for ZookeeperRegistryCenter.persist with utf-8

yangl:


terrymanu:
thank you for code contribute

Issue #312, opened 2017-05-10
Enhancement:Automatic reconciliation


hanahmily:
The `elastic-job-cloud-scheduler` provides an API for manual reconciliation. 

Since most mesos frameworks use automatic reconciliation, this feature should be added to `elastic-job-cloud-scheduler` to reduce manual intervention. 

Since automatic reconciliation is periodic, the original API is also retained to handle emergencies

PR #313, opened 2017-05-10
Fix for #312

hanahmily:


haocao:
4. Disable reconcile by set reconcile_interval_minutes to -1.

hanahmily:
@haocao Enable reconcile by setting reconcile_interval_minutes more than 0. If set reconcile_interval_minutes to 0, reconcile service had been running before scheduler had registered to Mesos Master

PR #495, opened 2018-03-20
增加Spring Boot Starter和使用示例

yinjihuan:
增加Spring Boot Starter和使用示例


terrymanu:
elastic job is an official open-source project. we need all code at package io.elasticjob, such as package 'com.cxytiandi.job', we cannot merge it.

PR #552, opened 2018-07-14
fixed issue #472

cnn112002:
fixed issue #472

- add adapter to create LiteJobConfiguration for spring boot project
-
-


terrymanu (changes requested):
Sorry to reply late.
I cannot understand about you add the 2 classes e`SpringFlowJobConfigurationAdapter` and `SpringSimpleJobConfigurationAdapter`

It seems there is no caller to call them. what is the useful of these classes?

PR #630, opened 2019-02-27
支持毫秒级别调度作业

BoulCheng:
### A New Feature: Job Of Millisecond Level.
- Supports scheduling jobs at the millisecond level, with jobs scheduled at intervals of a certain number of milliseconds.
- In addition, this function has been applied to the production environment of large-scale options and futures trading websites and it works perfectly well.

- **Usage**: You simply only replace the cron expression of the second parameter of the JobCoreConfiguration#newBuilder method with something like "1551276000000;500", with 1551276000000 before the semicolon as the number of milliseconds for the start time of the job, and 500 after the semicolon as intervals of milliseconds for scheduling jobs.

### 一个新特性：毫秒级别调度作业
- 支持毫秒级别调度作业，间隔一定毫秒数调度作业。
- 另外该功能已运用于大型期权期货交易网站的生产环境且效果很好。

- **用法**：仅仅需要把JobCoreConfiguration#newBuilder方法的第二个参数cron的cron表达式替换为类似"1551276000000;500”, 分号前面1551276000000为任务开始时间的毫秒数，分号后面500为每间隔500毫秒调度一次作业.

### DIFF
--- a/elastic-job-lite-core/src/main/java/io/elasticjob/lite/internal/schedule/JobScheduleController.java
+++ b/elastic-job-lite-core/src/main/java/io/elasticjob/lite/internal/schedule/JobScheduleController.java
@@ -19,14 +19,9 @@ package io.elasticjob.lite.internal.schedule;
 
 import io.elasticjob.lite.exception.JobSystemException;
 import lombok.RequiredArgsConstructor;
-import org.quartz.CronScheduleBuilder;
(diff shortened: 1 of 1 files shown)


terrymanu (changes requested):
I don't think we need to milliseconds schedule in the distributed framework. The network IO may milliseconds latency.

PR #688, opened 2019-10-21
Upgrade quartz version from 2.2.1 to 2.3.0

### LINKED ISSUE #680: Upgrade quartz's version
We are using quartz 2.2.1 which is using C3P0. The license of C3P0 is LGPL which is not friendly for Apache license.

We should Upgrade quartz's version to 2.3.0 at least that using DBCP instead of C3P0 and is friendly for Apache license.

compilerduck:
Fixes #680: Upgrade quartz version from 2.2.1 to 2.3.0. Examples of Java, Spring, SpringBoot have been tested successfully, including all of three job types. No fails found.
-
-
-


pom.xml
```diff
@@ -93,7 +93,7 @@
                 <version>${quartz.version}</version>
                 <exclusions>
                     <exclusion>
-                        <groupId>c3p0</groupId>
+                        <groupId>com.mchange</groupId>
```
terrymanu, line 96: Can we remove c3p0 when quartz upgrade?

terrymanu (changes requested):

pom.xml
compilerduck: quartz 2.3.0 has two default implementations of pooling connection provider: c3p0 and HikaryCP, so if we don't use c3p0 in ElasticJob, we can safely remove the dependency from ElasticJob pom. So quartz does have a compile dependency to c3p0, but not mendatory for runtime.

terrymanu: The license of c3p0 is LGPL which is not friendly for Apache license, please use HikariCP instead of it

compilerduck: I'm a bit confused. Did you mean we are using quartz persistence in ElasticJob, which can persist trigger info to db? But actually, ElasticJob doesn't use quartz persistence capability, therefore c3p0 never exist in our pom at all (by adding <exclusion> under quartz dependency node in pom.xml, as seen in above code). This means we don't need use HikariCP currently, unless we add trigger persistence feature later.

compilerduck: Quartz 2.2.1 only provides a c3p0 implementation of pooling connection provider, without capability of customized extension. While in 2.3.0, a 'PoolingConnectionProvider' interface and two implementations (c3p0 and HicariCP) are provided, so we have choices in case we need store trigger info.

elastic-job-lite-spring/src/main/resources/META-INF/namespace/job.xsd
```diff
@@ -38,7 +38,7 @@
                 <xsd:attribute name="monitor-port" type="xsd:string" default="-1"/>
                 <xsd:attribute name="max-time-diff-seconds" type="xsd:string" default="-1"/>
                 <xsd:attribute name="failover" type="xsd:string" default="false"/>
-                <xsd:attribute name="reconcile-interval-minutes" type="xsd:int" default="10"/>
+                <xsd:attribute name="reconcile-interval-minutes" type="xsd:string" default="10"/>
```
terrymanu: Please use other pull request to change it. Let's make this pull request do upgrade quartz only.

pom.xml
terrymanu: No, we do not need use c3p0, and do not need persist job into database.
So we should remove c3p0's dependency.
Please use HikariCP to instead of c3p0 for license reason

terrymanu (changes requested):

PR #745, opened 2020-06-05
Rename groupId to asf

### LINKED ISSUE #737: Change the base package , maven group and  maven artifactId
Since the elastic-job-lite is transferred here ,  we should change the base package , maven group and  maven artifactId

terrymanu:
Fixes #737.


lovepoem:
Good job

PR #776, opened 2020-06-14
remove comment code

CYXK:
Fixes #ISSUSE_ID.

- git can help us manage code, so comment code should remove.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -311,17 +311,6 @@
                     <artifactId>maven-jar-plugin</artifactId>
                     <version>${maven-jar-plugin.version}</version>
                 </plugin>
-                <!--<plugin>-->
-                    <!--<groupId>org.apache.maven.plugins</groupId>-->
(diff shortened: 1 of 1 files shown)


terrymanu (changes requested):
Sorry, this plugin cannot be remove because I just comment it temporary and make the test pass, I will investigate why it cause test cases broken.

Technoboy-:
> Sorry, this plugin cannot be remove because I just comment it temporary and make the test pass, I will investigate why it cause test cases broken.

Hi, John. I try to run the command with 'mvn clean install' after un-comment the plugin 'maven-surefire-plugin', it shows BUILD SUCCESS . I think we don't need to comment it. Please try again.  If test case can't pass, please let me know, I will fix it.

terrymanu:
@Technoboy- Thank you for reminder, I just fix it at #779

PR #781, opened 2020-06-17
Remove instance node in register server when shutdown (#780)

### LINKED ISSUE #780: instance node in register center could't be removed when procedure  shutdown(using spring)
I found the instance node in register center could't be removed when the procedure shutdown

the master branch version
ElasticJob-Lite
### Expected behavior
instance node in register server(zk) be removed
### Actual behavior
instance node still in zk until the heartbeat timeout
the shutdown method `schedulerFacade.shutdownInstance()` removed the instance in memory and didn't remove instance node when spring desory .

it remove the regCenter also,  so in the next  JobShutdownHookPlugin shutdown step,method  `new InstanceService(regCenter, jobName).removeInstance()` could not being called and the jobinstance node keep alive in the register center .

1. init the job server using spring
2. gracefully shutdown the jvm process

mmx1960:
Fixes #780.

- remove instance node in register server when shutdown


elastic-job-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/schedule/JobRegistry.java
```diff
@@ -167,6 +168,8 @@ public void shutdown(final String jobName) {
         CoordinatorRegistryCenter regCenter = regCenterMap.remove(jobName);
         if (null != regCenter) {
             regCenter.evictCacheData("/" + jobName);
+            //fix #780 remove instance in register server , place here ensure job `isShutdown`
```
Technoboy-: No need to comment this.

Technoboy-:
JobShutdownHookPlugin#shutdown has removed the instance. Please take a look and confirm .

Technoboy-:
First, I would like to describe how to reproduce this problem. 
```
1. Using spring context to start the application with calling classPathXmlApplicationContext.registerShutdownHook();
If not call this method, the issue could be reproduced.
2. Then try to shutdown the application.
3. List the zk path, you will see the children nodes of instances are not deleted.
ls /namespace/jobname/instances
```
Second, why does call registerShutdownHook will not delete instance children nodes?
Because, if application registered spring shutdown hook, it will call shutdown method of JobRegistry and remove CoordinatorRegistryCenter before JobShutdownHookPlugin#shutdown. That means it won't call 'new InstanceService(regCenter, jobName).removeInstance();' in the  JobShutdownHookPlugin#shutdown.
Last, I think this pr is an improvement. For current implementation, children nodes of instances are EPHEMERAL,  as the process shutdown,  even though it's not deleted actively, but it can be removed by zookeeper due to session timeout.
Please check .

mmx1960:
> First, I would like to describe how to reproduce this problem.
> 
> ```
> 1. Using spring context to start the application with calling classPathXmlApplicationContext.registerShutdownHook();
> If not call this method, the issue could be reproduced.
> 2. Then try to shutdown the application.
> 3. List the zk path, you will see the children nodes of instances are not deleted.
> ls /namespace/jobname/instances
> ```
> 
> Second, why does call registerShutdownHook will not delete instance children nodes?
> Because, if application registered spring shutdown hook, i will call shutdown method of JobRegistry and remove CoordinatorRegistryCenter before JobShutdownHookPlugin#shutdown. That means it won't call 'new InstanceService(regCenter, jobName).removeInstance();' in the JobShutdownHookPlugin#shutdown.
> Last, I think this pr is an improvement. For current implementation, children nodes of instances are EPHEMERAL, as the process shutdown, even though it's not deleted actively, but it can be removed by zookeeper due to session timeout.
> Please check .

great, thanks for review and describe more clearer, that's what i mean, this pr is the compensate when spring shutdown hook and quartz shutdown hook effected both

Technoboy-:
After talking with @terrymanu, we conclude that this pr can resolve the issue. But the job can not shutdown gracefully.  So John suggests to do more works on graceful shutdown to resolve this problem. This is really much more better. Please check @mmx1960 , and looking forward to your pr.

mmx1960:
optimize at #798

PR #864, opened 2020-06-29
update some version

### LINKED ISSUE #852: Will update spring version?
I notice that the newest version of spring is 5.0+, but ejob use version is 4.0 +。So will you have a plan,that , import the lastest spring and springBoot project ?

Mark-WJQ:
Fixes #852 .

1. import spring-boot-dependencies in dependencyManagement with the scope of import,so we can use springboot in sub module,the version is 2.3.1.RELEASE.
2. unpdate spring version range [3.1.0.RELEASE,5.0.0.M1) to [3.1.0.RELEASE,5.0.0.M1)
3. update lombok version 1.16.4 to 1.18.12
4. update aspectj.version 1.8.0 to 1.9.1
5. update logback.version 1.1.2 to 1.2.3
6. I run the comman 'mvn clean install -U -T 4',it's ok,all tests are pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,17 +43,17 @@
         <maven.version.range>[3.0.4,)</maven.version.range>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.locale>zh_CN</project.build.locale>
-
-        <springframework.version>[3.1.0.RELEASE,5.0.0.M1)</springframework.version>
-
+        
+        <springframework.version>[3.1.0.RELEASE,5.2.7.RELEASE]</springframework.version>
+        <springboot.version>2.3.1.RELEASE</springboot.version>
         <guava.version>18.0</guava.version>
         <commons-lang3.version>3.4</commons-lang3.version>
         <quartz.version>2.3.0</quartz.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -43,17 +43,17 @@
         <maven.version.range>[3.0.4,)</maven.version.range>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.locale>zh_CN</project.build.locale>
-
-        <springframework.version>[3.1.0.RELEASE,5.0.0.M1)</springframework.version>
-
+        
+        <springframework.version>[3.1.0.RELEASE,5.2.7.RELEASE]</springframework.version>
```
terrymanu: If spring framework upgrade to new version, we still need to modify here,
can we keep the upper boundary open?

terrymanu (changes requested):

pom.xml
Mark-WJQ: I get it, how do you think '<springframework.version>[3.1.0.RELEASE,)</springframework.version>' ?

terrymanu: Great

Issue #792, opened 2020-06-19
New job API


terrymanu:
New job API

- [X] Decouple JobItemExecutor and JobExecutor
- [x] Add job instance into JobScheduler API
- [x] Remove class tag for spring namespace
- [x] Remove class name for job configuration
- [x] Simplify Job API level, merge configuration object to single
- [x] Remove static job type and use SPI to introduce job type

PR #917, opened 2020-07-03
Refactor TypedJob

terrymanu:
For #792.


terrymanu:
Closed because invalid commit version

PR #977, opened 2020-07-06
fix the issue that the elastic job console encode the password with Base64, which is not safety!

luky116:
Fixes #ISSUSE_ID.

- change the Basic HTTP Authentication  to Digest HTTP Authentication


terrymanu (changes requested):
Could you fix the CI error?

luky116:
> Could you fix the CI error?

ok,i will recommit again

PR #981, opened 2020-07-08
Fix the demo which doesn't work properly.

TeslaCN:

* Correct Spring version
* Exclude HikariCP from Quartz
* Replace @Resource with @Autowired


examples/elastic-job-example-lite-springboot/src/main/java/org/apache/shardingsphere/elasticjob/lite/example/config/DataflowJobConfig.java
```diff
@@ -36,7 +37,7 @@
     @Resource
     private ZookeeperRegistryCenter regCenter;
     
-    @Resource
+    @Autowired
```
terrymanu: What is the different with @Resource and @Autowired?

terrymanu (approved):

PR #1010, opened 2020-07-09
Add Spring Boot Starter

### LINKED ISSUE #970: Add spring boot starter
It is better to add spring boot starter for ElasticJob.

TeslaCN:
Fixes #970.

- Add document for ElasticJob Spring Boot Starter.
- Provide a user-friendly Spring Boot Starter.
- Provide a simple demo for getting started.


terrymanu (changes requested):
The YAML is different with configuration which persisted with reg center, do we need to keep 2 YAML format?

PR #1060, opened 2020-07-14
 Upgrade guava and curator version to newest

terrymanu:
Upgrade curator to version 5.1.0, ZooKeeper only support 3.6 + for now;
Upgrade guava to version 29.0-jre


rudy2steiner:
Greate!

PR #1082, opened 2020-07-16
CloudJobConfigurationPOJO instead of CloudJobConfiguration

### LINKED ISSUE #1047: Use CloudJobConfigurationPOJO instead of CloudJobConfiguration for CloudJobRestfulApi
We are using CloudJobConfiguration to exchange message from frontend and backend.
The class `JobConfigurationGsonTypeAdapter` is adapter with object and JSON string.
It is better to remove class `JobConfigurationGsonTypeAdapter`, and reuse CloudJobConfigurationPOJO to exchange message.
The tasks are:

- [x] Use CloudJobConfigurationPOJO instead of CloudJobConfiguration in CloudJobRestfulApi
- [x] Adjust JavaScript of frontend
- [x] Remove `GsonFactory.registerTypeAdapter(CloudJobConfiguration.class, new JobConfigurationGsonTypeAdapter())` from CloudJobRestfulApi.init()
- [x] Remove class `JobConfigurationGsonTypeAdapter`

rudy2steiner:
Fixes #1047.

- CloudJobConfigurationPOJO instead of CloudJobConfiguration
- Remove useless JobConfigurationGsonTypeAdapter and it's register code
- BootstrapEnvironment load bug fix
    Make sure conf/elastic-job-cloud-scheduler.properties can be loaded in dev and excluded in the jar, so the user could use 
   conf/elastic-job-cloud-scheduler.properties on the classpath to override env properties in production

As I know, there is no need to change any frontend code


elasticjob-cloud/elasticjob-cloud-scheduler/pom.xml
```diff
@@ -142,12 +142,21 @@
...
-                   <exclude>conf/*</exclude>
                    <exclude>assembly/*</exclude>
                </excludes>
             </resource>
         </resources>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.3.2</version>
```
terrymanu: Please use dependency management to manage version

terrymanu (changes requested):

elasticjob-cloud/elasticjob-cloud-scheduler/pom.xml
rudy2steiner: Ok

terrymanu (approved):

PR #1171, opened 2020-07-19
Use spring boot instead of jetty for cloud console

### LINKED ISSUE #1050: Use spring boot instead of jetty for cloud console
The console of ElasticJob Lite has already use spring boot, but the cloud console is still use jetty.
We need to keep consist with them.
Please Use spring boot instead of `RestfulServer` in ElasticJob Cloud.

We need call spring boot at `SchedulerService.start()`, replace of `restfulService.start()`

kekerzzZ:
Fixes #1050.


terrymanu (changes requested):
Could you add test cases?
The test coverage decreased 4.2%

menghaoranss:
There are still some jersery dependencies, could you please check again and fix them.

kekerzzZ:
ok. I will fix it.

menghaoranss (approved):

PR #1329, opened 2020-08-03
修复哈希表初始容量以避免发生扩容

BoulCheng:
修复哈希表初始容量以避免发生扩容


terrymanu:
Why not use sharding total count as initial capacity of hash map?

BoulCheng:
> Why not use sharding total count as initial capacity of hash map?

Because the key of the hash table is the job instance. If the shardingTotalCount is larger than the total number of job instances, space will be wasted. If the shardingTotalCount is smaller than the total number of job instances, the hash table will resize.

terrymanu (approved):

PR #1335, opened 2020-08-05
misfire depends on monitorExecution

BoulCheng:
misfire depends on monitorExecution. If monitorExecution flag is false then misfire will fail even if misfire flag is true, because a job running node can only be created if monitorExecution is marked true.


terrymanu (changes requested):
It is better to do not change user configuration, it will make user confuse.
We can add the hint in to document

terrymanu:
I update the document at #1336.
And this pr is wrong because of misfire is no relationship with execution monitor.

BoulCheng:
> It is better to do not change user configuration, it will make user confuse.
> We can add the hint in to document

OK

PR #1337, opened 2020-08-06
fix thread pools execute jobs uniformly

BoulCheng:
Thread pools execute jobs uniformly


terrymanu (changes requested):
The change is  not correct because if only one item assigned, it is unnecessary to create new thread to process it.

terrymanu:
I notice your pull request is wrong sometimes.
Can you discuss in issue before submit the pull request?
It is better to discuss first.

BoulCheng:
> The change is not correct because if only one item assigned, it is unnecessary to create new thread to process it.

After JobBootstrap#shutdown() is called, The job is still executed if only one item assigned, there may still need to add a judgment on whether the thread pool is shutdown or not.

PR #1375, opened 2020-08-15
[optimization][cloud]code optimization

CalvinKirs:
code optimization

### DIFF
--- a/elasticjob-cloud/elasticjob-cloud-scheduler/src/main/java/org/apache/shardingsphere/elasticjob/cloud/scheduler/producer/TransientProducerRepository.java
+++ b/elasticjob-cloud/elasticjob-cloud-scheduler/src/main/java/org/apache/shardingsphere/elasticjob/cloud/scheduler/producer/TransientProducerRepository.java
@@ -33,7 +33,7 @@ final class TransientProducerRepository {
     
     synchronized void put(final JobKey jobKey, final String jobName) {
         remove(jobName);
-        List<String> taskList = cronTasks.computeIfAbsent(jobKey, k -> Collections.singletonList(jobName));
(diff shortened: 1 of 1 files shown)


elasticjob-cloud/elasticjob-cloud-scheduler/src/main/java/org/apache/shardingsphere/elasticjob/cloud/scheduler/producer/TransientProducerRepository.java
```diff
@@ -34,16 +33,7 @@
...
-        if (null == taskList) {
-            taskList = new CopyOnWriteArrayList<>();
-            taskList.add(jobName);
-            cronTasks.put(jobKey, taskList);
-            return;
-        }
-        if (!taskList.contains(jobName)) {
-            taskList.add(jobName);
-        }
+        cronTasks.computeIfAbsent(jobKey, k -> Collections.singletonList(jobName));
```
terrymanu, line 36: This change maybe incorrect because the `cronTasks` may include multiple job names, not for singletonList only

terrymanu (changes requested):

elasticjob-cloud/elasticjob-cloud-scheduler/src/main/java/org/apache/shardingsphere/elasticjob/cloud/scheduler/producer/TransientProducerRepository.java
CalvinKirs: Thank you very much for your review, so the previous code should be missing, right? On line 45, it is not updated

PR #1382, opened 2020-08-17
[Improvement][common]get the ip configured by hostname first

CalvinKirs:
When obtaining the ip, the ip configured by the hostname should be obtained first. At the same time, if the user configures the ip corresponding to the hostname, there is no need to do a connectivity test at this time

### DIFF
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/env/IpUtils.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/env/IpUtils.java
@@ -26,7 +26,11 @@ import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.net.UnknownHostException;
 import java.net.Inet6Address;
-import java.util.*;
+import java.util.Enumeration;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
 
 /**
  * IP address utility.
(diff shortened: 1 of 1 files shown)


Technoboy-:
On the face of it, InetAddress.getLocalHost() should give you the IP address of this host. The problem is that a host could have lots of network interfaces, and an interface could be bound to more than one IP address. And to top that, not all IP addresses will be reachable outside of your machine or LAN. For example, they could be IP addresses for virtual network devices, private network IP addresses, and so on.
What this means is that the IP address returned by InetAddress.getLocalHost() might not be the right one to use.
So the current implementation looks nice.
BTW, reference to [Code of Conduct](https://shardingsphere.apache.org/community/en/contribute/code-conduct/) for pull request.

CalvinKirs:
Thank you very much for your answers, I can think my thinking is not comprehensive enough.  I will sort out the problem later.  Then close it.




| |
ACM_China
|
|
邮箱：acm_china@yeah.net
|

Signature is customized by Netease Mail Master

On 08/18/2020 13:12, Tboy wrote:

On the face of it, InetAddress.getLocalHost() should give you the IP address of this host. The problem is that a host could have lots of network interfaces, and an interface could be bound to more than one IP address. And to top that, not all IP addresses will be reachable outside of your machine or LAN. For example, they could be IP addresses for virtual network devices, private network IP addresses, and so on.
What this means is that the IP address returned by InetAddress.getLocalHost() might not be the right one to use.

CalvinKirs:
In the production environment, especially docker, our ip is often dynamic, so this method is not recommended. In addition, it is better to test the connectivity of the hostip configured by the user. This is an anti-dementia design to avoid users setting useless ip

CalvinKirs:
In the production environment, especially docker, our ip is often dynamic, so this method is not recommended. In addition, it is better to test the connectivity of the hostip configured by the user. This is an anti-dementia design to avoid users setting useless ip

PR #1432, opened 2020-08-31
Improve HTTP job executor

### LINKED ISSUE #1431: HTTP job executor: Retrieve response body rather than discard it when response is not OK
When using HTTP job executor, if response code is not 200, executor will throw an exception. 

Log the response body in WARN level if HTTP request is not OK.

TeslaCN:
Fixes #1431 

- Log the response body in WARN level if HTTP request is not OK.
- Replace explicit close with try-with-resource.
- Adjust dependencies scope and add optional logger dependency for test logging.

### DIFF
--- a/elasticjob-executor/elasticjob-executor-type/elasticjob-http-executor/src/main/java/org/apache/shardingsphere/elasticjob/http/executor/HttpJobExecutor.java
+++ b/elasticjob-executor/elasticjob-executor-type/elasticjob-http-executor/src/main/java/org/apache/shardingsphere/elasticjob/http/executor/HttpJobExecutor.java
@@ -31,9 +31,10 @@ import org.apache.shardingsphere.elasticjob.infra.exception.JobExecutionExceptio
 import org.apache.shardingsphere.elasticjob.infra.json.GsonFactory;
 
 import java.io.BufferedReader;
-import java.io.DataOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


elasticjob-executor/elasticjob-executor-type/elasticjob-http-executor/src/main/java/org/apache/shardingsphere/elasticjob/http/executor/HttpJobExecutor.java
```diff
@@ -116,6 +118,10 @@ private boolean isWriteMethod(final String method) {
         return Arrays.asList("POST", "PUT", "DELETE").contains(method.toUpperCase());
     }
     
+    private boolean isRequestSucceed(final int httpStatusCode) {
+        return HttpURLConnection.HTTP_BAD_REQUEST > httpStatusCode;
```
Technoboy-, line 122: Why not use 'HttpURLConnection.HTTP_OK == httpStatusCode' to accurate the result ?

TeslaCN: This condition is refer to `sun.net.www.protocol.http.HttpURLConnection#getErrorStream`, which implemented `java.net.HttpURLConnection`.

Technoboy- (approved):

PR #1467, opened 2020-09-19
Move ElasticJobListener to JobConfiguration

### LINKED ISSUE #1451: Move ElasticJobListener to JobConfiguration
ElasticJobListener is a bit heavy as a schedule parameter., it's better to configure in JobConfiguration like ErrorHandler.
So JobListener should listener shard item or job level event?

Technoboy-:
Fixes #1451.


terrymanu (changes requested):
Cloud we consider about use String to describe listener type?
Maybe keep consist with jobErrorHandlerType?

PR #1501, opened 2020-09-25
Add Pre-Check for param `elasticJob` in JobScheduler's constructor 

### LINKED ISSUE #1496: Confusing exception information when  construct `JobScheduler`
If  param  `elasticJob` is null, the following `JobScheduler` contructor will throw **JobConfigurationException:  Can not find executor for elastic job type `null`**

```java
// at line 78
public JobScheduler(final CoordinatorRegistryCenter regCenter,
                    final ElasticJob elasticJob,
                    final JobConfiguration jobConfig,
                    final TracingConfiguration<?> tracingConfig,
                    final ElasticJobListener... elasticJobListeners)
```

this constructor don't care param `String elasticJobType`, so the error message  make user confusing. 

It's caused by the following code segment

```java
// at line 78
public JobScheduler(...) {
    // in this constructor,  elasticJobType is always null
    elasticJobType = null; // at line 82
    // elasticJob is null, will call `ElasticJobExecutor` constructor, then throw JobConfigurationException
    jobExecutor = null == elasticJob ? new ElasticJobExecutor(elasticJobType, jobConfig, jobFacade) : new ElasticJobExecutor(elasticJob, jobConfig, jobFacade); // at line 87
}
```

so, i think

- should do Pre-check for param `elasticJob`
- give more clearly error message

could i improve it ?

vran-dev:
Fixes #1496  

-  Do Pre-check for param elasticJob
-  Give more clearly error message

### DIFF
--- a/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/schedule/JobScheduler.java
+++ b/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/schedule/JobScheduler.java
@@ -41,6 +41,7 @@ import org.quartz.simpl.SimpleThreadPool;
 import java.util.Collection;
 import java.util.Properties;
 import java.util.stream.Collectors;
+import com.google.common.base.Preconditions;
 
 /**
  * Job scheduler.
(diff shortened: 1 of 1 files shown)


elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/schedule/JobScheduler.java
```diff
@@ -53,7 +54,7 @@
     private final CoordinatorRegistryCenter regCenter;
     
     private final String elasticJobType;
-
```
TeslaCN: Keep indents (including blank lines) consistent with the previous one.
Refer to https://shardingsphere.apache.org/community/en/contribute/code-conduct/

vran-dev: thx.

terrymanu (changes requested):
This pr is incorrect.
We need to permit elasticJob is null because user may use elasticJobType as param

For example: 
Simple job and dataflow job use elasticJob for user customized;
Script and HTTP job do not need user development anything, just provide the script shell or HTTP URL, so they use elasticJobType.

PR #1508, opened 2020-09-28
Distinguish guarantee service by job name in singleton AbstractDistributeOnceElasticJobListener

### LINKED ISSUE #1502: Unexpected behaviour may happen if different jobs use same distribute once jobListenerType
There is a stateful field 'guaranteeService' in AbstractDistributeOnceElasticJobListener. The SPI will create only one instance of distribute once listener. If more than a job refer the same jobListenerType, the 'guaranteeService' would be overwrite and something unexpected may happen.

TeslaCN:
Fixes #1502 .

- Distinguish guarantee service by job name in singleton AbstractDistributeOnceElasticJobListener.
- Setup listener's parameters in a format like query string.


TeslaCN:
I think we can configure timeout of AbstractDistributeOnceListener like the following code.

```java
JobConfiguration jobConfiguration = JobConfiguration.newBuilder("test", 2)
    .jobListenerTypes("simpleListener", "distributeListener?startedTimeoutMilliseconds=1000&completedTimeoutMilliseconds=2000").build();
```

terrymanu (approved):

PR #1525, opened 2020-09-29
Use java properties to config error handler's configuration

### LINKED ISSUE #1516: Use java properties to config error handler's configuration
We are using YAML file to config the error handler's configuration which is not keep consist with original elasticjob config.
It is better to use java configuration only. YAML just one of way to `core config`.

- [ ] error-handler-email.yaml
- [ ] error-handler-wechat.yaml
- [ ] error-handler-dingtalk.yaml

gzdzss:
Fixes #1516.

- Use java properties to config error handler's configuration


TeslaCN (commented):
JobErrorHandler is singleton. What if there are more than one job and their properties are different?

lovepoem:
Please resolve the conflicts  @gzdzss

PR #1542, opened 2020-10-03
Correct hyperlinks in GitHub templates

TeslaCN:


.github/workflows/maven.yml
```diff
@@ -29,7 +29,7 @@ on:
 jobs:
   build:
 
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
```
Technoboy-, line 32: I'm not sure about this. Shardingsphere uses macos.

Technoboy-:
cc @terrymanu

.github/workflows/maven.yml
TeslaCN: I think Linux is more common in actual environment and it is a little faster than macOS in GitHub Action.

TeslaCN:
I'm going to separate the commits in different PRs.

PR #1721, opened 2020-11-13
#1720 Let user can customize the jobInstanceId

### LINKED ISSUE #1720: Customize the JobInstance
## Feature Request

For some reason, we need to customize the JobInstance ID, so that we can assign specific shard number to specific JobInstance.

And as I found in JobInstance.java, the JobInstance ID is fixed to IP concate with JVM name. The code as following:

```java
    public JobInstance() {
        jobInstanceId = IpUtils.getIp() + DELIMITER + ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
    }

```

Yes, I cannot identify JobInstance in JobShardingStrategy 
Suggest to check whether a "jobInstanceId" property is exists, if so, use IP concat with this property instead of jvm name.
code like this:
```java
    public JobInstance() {
        String jobInstIdSuffix = System.getProperty("jobInstanceId");
        if (jobInstIdSuffix == null) {
        	jobInstIdSuffix = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
        }
	jobInstanceId = IpUtils.getIp() + DELIMITER + jobInstIdSuffix;
    }
```
BTW, changed constances to public, so that we can cut out the ID suffix, e.g. 10.0.0.110@-@001 to cut out "001"

johnny2002:
Fixes #1720 .

Detect System Property "jobInstanceId", if it is exists, use IP concat with this property instead of jvm name.
-
-

### DIFF
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
@@ -39,11 +39,11 @@ public final class JobInstance {
     private final String jobInstanceId;
     
     public JobInstance() {
-        String jobInstanceIdSuffix  = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
-        if (null == jobInstanceIdSuffix ) {
-            jobInstanceIdSuffix  = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
+        String jobInstanceIdSuffix = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
+        if (null == jobInstanceIdSuffix) {
+            jobInstanceIdSuffix = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
         }
-        jobInstanceId = IpUtils.getIp() + DELIMITER + jobInstanceIdSuffix ;
+        jobInstanceId = IpUtils.getIp() + DELIMITER + jobInstanceIdSuffix;
     }
     
     /**
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
@@ -34,12 +34,12 @@ public final class JobInstance {
     
     private static final String DELIMITER = "@-@";
     
-    private static final String PARAM_JOB_INST_ID = "jobInstanceId";
+    private static final String PROPERTY_JOB_INSTANCE_ID = "jobInstanceId";
 
     private final String jobInstanceId;
     
     public JobInstance() {
-        String jobInstIdSuffix = System.getProperty(PARAM_JOB_INST_ID);
+        String jobInstIdSuffix = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
         if (jobInstIdSuffix == null) {
             jobInstIdSuffix = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
         }
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
@@ -40,7 +40,7 @@ public final class JobInstance {
     
     public JobInstance() {
         String jobInstIdSuffix = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
-        if (jobInstIdSuffix == null) {
+        if (null == jobInstIdSuffix) {
             jobInstIdSuffix = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
         }
         jobInstanceId = IpUtils.getIp() + DELIMITER + jobInstIdSuffix;
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
@@ -32,9 +32,9 @@ import java.lang.management.ManagementFactory;
 @EqualsAndHashCode(of = "jobInstanceId")
 public final class JobInstance {
     
-    private static final String DELIMITER = "@-@";
+    public static final String DELIMITER = "@-@";
     
-    private static final String PROPERTY_JOB_INSTANCE_ID = "jobInstanceId";
+    public static final String PROPERTY_JOB_INSTANCE_ID = "jobInstanceId";
 
     private final String jobInstanceId;
     
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
@@ -39,11 +39,11 @@ public final class JobInstance {
     private final String jobInstanceId;
     
     public JobInstance() {
-        String jobInstIdSuffix = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
-        if (null == jobInstIdSuffix) {
-            jobInstIdSuffix = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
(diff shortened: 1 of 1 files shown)


elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
```diff
@@ -34,10 +34,16 @@
     
     private static final String DELIMITER = "@-@";
     
+    private static final String PARAM_JOB_INST_ID = "jobInstanceId";
```
TeslaCN: Name it `PROPERTY_JOB_INSTANCE_ID` may be better.

elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
```diff
@@ -34,10 +34,16 @@
...
+
     private final String jobInstanceId;
     
     public JobInstance() {
-        jobInstanceId = IpUtils.getIp() + DELIMITER + ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
+        String jobInstIdSuffix = System.getProperty(PARAM_JOB_INST_ID);
+        if (jobInstIdSuffix == null) {
```
TeslaCN: Constant is on the left.

```suggestion
        if (null == jobInstIdSuffix) {
```

TeslaCN (changes requested):

elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/handler/sharding/JobInstance.java
```diff
@@ -34,10 +34,16 @@
     
     private static final String DELIMITER = "@-@";
     
+    private static final String PROPERTY_JOB_INSTANCE_ID = "jobInstanceId";
+
     private final String jobInstanceId;
     
     public JobInstance() {
-        jobInstanceId = IpUtils.getIp() + DELIMITER + ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
+        String jobInstIdSuffix = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
```
TeslaCN: It's better to keep the names consistent.
```suggestion
        String jobInstanceIdSuffix = System.getProperty(PROPERTY_JOB_INSTANCE_ID);
```

TeslaCN:
Please fix the checkstyle.

johnny2002:
BTW, changed  constances to public, so that we can cut out the ID suffix, e.g. 10.0.0.110@-@001 to cut out "001"

Technoboy-:
Hi, we discuss this pr and find that it may be no need to change it. We think the original way is enough and simple.

terrymanu (changes requested):
The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.

johnny2002:
> The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.

if we don't make this changes,  then there is no way to fullfill our requirement.  our requirement is that we want to assign specific partition to specific job instance. 
we need to customize the JobInstance class,  while noway found. the best way is to provide a  mechanism so that developers can extend the JobInstance class

TeslaCN:
> > The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.
> 
> if we don't make this changes, then there is no way to fullfill our requirement. our requirement is that we want to assign specific partition to specific job instance.
> we need to customize the JobInstance class, while noway found.

Sorry to take up your time. I realized that the jobInstanceId is an internal implementation of ElasticJob and it is not recommended to expose to user. Could you consider defining your own instanceId in somewhere?

johnny2002:
> > > The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.
> > 
> > 
> > if we don't make this changes, then there is no way to fullfill our requirement. our requirement is that we want to assign specific partition to specific job instance.
> > we need to customize the JobInstance class, while noway found.
> 
> Sorry to take up your time. I realized that the jobInstanceId is an internal implementation of ElasticJob and it is not recommended to expose to user. Could you consider defining your own instanceId in somewhere?

but I need to know the JobInstance ID in sharing strategy implementation. I cannot find out a way, can you?

johnny2002:
> > > The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.
> > 
> > 
> > if we don't make this changes, then there is no way to fullfill our requirement. our requirement is that we want to assign specific partition to specific job instance.
> > we need to customize the JobInstance class, while noway found.
> 
> Sorry to take up your time. I realized that the jobInstanceId is an internal implementation of ElasticJob and it is not recommended to expose to user. Could you consider defining your own instanceId in somewhere?

as a matter of fact,  user can see the job instance id in the console. and most of all, user can only distinguish JobInstances through id. in this way, you can't say it is an internal feature

TeslaCN:
> > > > The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.
> > > 
> > > 
> > > if we don't make this changes, then there is no way to fullfill our requirement. our requirement is that we want to assign specific partition to specific job instance.
> > > we need to customize the JobInstance class, while noway found.
> > 
> > 
> > Sorry to take up your time. I realized that the jobInstanceId is an internal implementation of ElasticJob and it is not recommended to expose to user. Could you consider defining your own instanceId in somewhere?
> 
> but I need to know the JobInstance ID in sharing strategy implementation. I cannot find out a way, can you?

May I know some details about your implementation of JobShardingStrategy? Or have you tried implementing your own JobItemExecutor?

johnny2002:
> > > > > The proposal of this pr still need to be discussed, I post my comment in the issue, we can talk about it there.
> > > > 
> > > > 
> > > > if we don't make this changes, then there is no way to fullfill our requirement. our requirement is that we want to assign specific partition to specific job instance.
> > > > we need to customize the JobInstance class, while noway found.
> > > 
> > > 
> > > Sorry to take up your time. I realized that the jobInstanceId is an internal implementation of ElasticJob and it is not recommended to expose to user. Could you consider defining your own instanceId in somewhere?
> > 
> > 
> > but I need to know the JobInstance ID in sharing strategy implementation. I cannot find out a way, can you?
> 
> May I know some details about your implementation of JobShardingStrategy? Or have you tried implementing your own JobItemExecutor?

my sharding strategy is that partition 1 assign to JobInstance1, partition 2 assign to JobInstance 2,  and so on. I will show you the sample code later

johnny2002:
samlple code:


```java
/**
 * @author WeiZhou
 * 
 * shardingTotalCount should be as same as JobInstances count.
 * Neighbor 2 JobInstances backup eachother.
 * Each partition will be assigned to same Jobinstance number or its backup.
 * shardingTotalCount和jobInstances数量要一样.
 * 相邻两个JobInstances互相备份.
 * 每个作业分片要分配给相同号码的JobInstance或是它的备份单元
 */
@Slf4j
public class UnitizedAllocationJobShardingStrategy implements JobShardingStrategy {
    int[] backupMap = new int[] { 2, 1, 4, 3, 6, 5, 8, 7 };//Unit 1<>2, 3<>4, 5<>6, 7<>8 backup eachother

    @Override
    public Map<JobInstance, List<Integer>> sharding(final List<JobInstance> jobInstances, final String jobName, final int shardingTotalCount) {
        if (jobInstances.isEmpty()) {
            return Collections.emptyMap();
        }
        JobInstance[] instArray = new JobInstance[shardingTotalCount];
        for (JobInstance jobInst : jobInstances) {
            //Job instance IDs like xxx@-@001, xxx@-@002, xxx@-@003 ...
            String jobInstanceId = jobInst.getJobInstanceId();
            jobInstanceId = jobInstanceId.substring(jobInstanceId.indexOf(JobInstance.DELIMITER) + JobInstance.DELIMITER.length());
            int jobInstNum = Integer.parseInt(jobInstanceId);
            instArray[jobInstNum - 1] = jobInst;
        }
        Map<JobInstance, List<Integer>> result = new LinkedHashMap<>(jobInstances.size(), 1);
        for (int i = 0; i < shardingTotalCount; i++) {
            //looup same number JobInstance
            JobInstance locatedJobInst = instArray[i];
            log.debug("master of {} is {}", i, locatedJobInst);
            if (locatedJobInst == null) {
                //lookup backup JobInstance
                if (backupMap[i] - 1 >= instArray.length || instArray[backupMap[i] - 1] == null) {
                    throw new RuntimeException("No main or backup job executor found for unit:" + (i + 1) + ", backup unit:" + backupMap[i]);
                }
                locatedJobInst = instArray[backupMap[i] - 1];
                log.debug("backup of {} is {}", i, locatedJobInst);
            }
                
            List<Integer> parts = result.get(locatedJobInst);
            if (parts == null) {
                parts = new ArrayList<>(2);
                result.put(locatedJobInst, parts);
            }
             parts.add(i);   
        }
        return result;
    }


    @Override
    public String getType() {
        return "UNIT_ALLOCATION";
    }
}

```

TeslaCN:
Close due to no conclusion at present.

PR #1734, opened 2020-11-25
Add null check in SnapshotService to avoid NPE

### LINKED ISSUE #1732: SnapshotService will be unavailable in specific situation
Try dumping a job exists in current namespace but not scheduled by the process which runs the SnapshotService.

### Actual behavior

```
Exception in thread "Thread-0" java.lang.NullPointerException
	at org.apache.shardingsphere.elasticjob.lite.internal.snapshot.SnapshotService.dumpDirectly(SnapshotService.java:117)
	at org.apache.shardingsphere.elasticjob.lite.internal.snapshot.SnapshotService.process(SnapshotService.java:103)
	at org.apache.shardingsphere.elasticjob.lite.internal.snapshot.SnapshotService.lambda$openSocket$0(SnapshotService.java:79)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

After that, the SnapshotService will be unavailable.

TeslaCN:
Fixes #1732 

- Add null check in SnapshotService to avoid NPE.

### DIFF
--- a/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/snapshot/SnapshotService.java
+++ b/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/snapshot/SnapshotService.java
@@ -115,8 +115,8 @@ public final class SnapshotService {
             CuratorCache cache = (CuratorCache) regCenter.getRawCache("/" + jobName);
             if (null != cache) {
                 Optional<ChildData> cacheData = cache.get(zkPath);
-                cachePath = cacheData.map(ChildData::getPath).orElse(zkPath);
-                cacheValue = cacheData.map(ChildData::getData).map(String::new).orElse(zkValue);
+                cachePath = cacheData.map(ChildData::getPath).orElse("");
+                cacheValue = cacheData.map(ChildData::getData).map(String::new).orElse("");
             }
             if (zkValue.equals(cacheValue) && zkPath.equals(cachePath)) {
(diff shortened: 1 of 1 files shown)


elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/snapshot/SnapshotService.java
```diff
@@ -109,14 +109,15 @@ private void process(final Socket socket) throws IOException {
...
-                zkValue = "";
-            }
+            String zkValue = Optional.ofNullable(regCenter.get(zkPath)).orElse("");
+            String cachePath = zkPath;
+            String cacheValue = zkValue;
             CuratorCache cache = (CuratorCache) regCenter.getRawCache("/" + jobName);
-            Optional<ChildData> cacheData = cache.get(zkPath);
-            String cachePath = cacheData.map(ChildData::getPath).orElse("");
-            String cacheValue = cacheData.map(childData -> new String(childData.getData())).orElse("");
+            if (null != cache) {
```
Technoboy-, line 116: If cache == null, that means cachePath equals zkPath. and will let the logic at line 121 always be true. it's not the same logic as before.

TeslaCN: `cache == null` means there is no cache, so there is no need to dump both zk data and cache data.

Technoboy- (approved):

PR #1774, opened 2020-12-13
Update GsonFactoryTest.java

y553546436:
Remove potential test order dependency

### DIFF
--- a/elasticjob-infra/elasticjob-infra-common/src/test/java/org/apache/shardingsphere/elasticjob/infra/json/GsonFactoryTest.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/test/java/org/apache/shardingsphere/elasticjob/infra/json/GsonFactoryTest.java
@@ -52,6 +52,19 @@ public final class GsonFactoryTest {
         });
         assertThat(beforeRegisterGson.toJson(new GsonFactoryTest()), is("{}"));
         assertThat(GsonFactory.getGson().toJson(new GsonFactoryTest()), is("test"));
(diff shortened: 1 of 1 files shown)


TeslaCN (changes requested):
Hi @y553546436 ,

I can't get your point. Could you clarify what the problem is?

y553546436:
> Hi @y553546436 ,
> 
> I can't get your point. Could you clarify what the problem is?

Hi @TeslaCN, this test modifies GsonFactory, which contains (static) fields shared among tests. If other tests are added later on which assume the default behavior of GsonFactory, they would fail. In another word, this test "pollutes" the shared program state and could potentially lead other tests run after it to fail. Though there is no severe problem now, it would be good to clean the state (remove the "pollution") at the end of the test.

TeslaCN:
Hi @y553546436 ,

I see what you mean now. But on the one hand, register another TypeAdapter cannot eliminate the impact, unless you remove the registered TypeAdapter from GsonBuilder. On the other hand, `GsonFactoryTest` is just a simple unit test class should not be used by other testcase.

TeslaCN:
Close due to no response.

y553546436:
Sorry that I didn't follow up because I did not receive the notification email from GitHub. 
Indeed, this PR is of poor quality because (1) it lacks good description, (2) it does not follow your PR template (one PR points to one issue), and (3) the code does not fully remove the pollution, as you have commented.
I sincerely apologize for my carelessness, and have opened a new PR that addresses the above three issues: #1901.

PR #1798, opened 2021-01-01
fix issue#1746

HombioChan:
Fixes #ISSUSE_1746.

- FullHttpResponse use default buffer(unpooled heap buffer)
- relase httpRequest refcnt in HandlerParameterDecoder


TeslaCN (changes requested):
The testcase `HandlerParameterDecoderTest` is using `EmbeddedChannel`. You need to set the request's reference count.

TeslaCN:
https://github.com/apache/shardingsphere-elasticjob/pull/1798/checks#step:5:1033
The assertion failed but the testcase passed because assertion is on another thread. We need further consideration about this issue.

PR #1864, opened 2021-03-31
add JobInstanceRegistry

Lucas-307:


sunkai-cai:
I can't get the point. Count you update the document？

PR #1885, opened 2021-04-28
#1883 fix performance loss for the misfired configuration that is not turned on

### LINKED ISSUE #1883: LiteJobFacade#isExecuteMisfired method is to first obtain the need for points from zk, and then determine whether misfired is turned on
## Feature Request

**The LiteJobFacade#isExecuteMisfired method first obtains whether points are needed from zk`(!isNeedSharding())`, and then judges whether misfired is enabled`configService.load(true).isMisfire()`. This is a performance loss for the misfired configuration that is not turned on.**

**just suggestion**

**Reverse order` !isNeedSharding() && configService.load(true).isMisfire()` -> `configService.load(true).isMisfire() && !isNeedSharding()`. According to the short-circuit characteristics, if the misfired configuration is not turned on, there will be no additional performance loss.**

seeger12138:
Fixes https://github.com/apache/shardingsphere-elasticjob/issues/1883

- Reverse order !isNeedSharding() && configService.load(true).isMisfire() -> configService.load(true).isMisfire() && !isNeedSharding(). According to the short-circuit characteristics, if the misfired configuration is not turned on, there will be no additional performance loss.


TeslaCN (approved):
Hi @seeger12138 
Thanks for your contribution!

PR #1901, opened 2021-06-07
Remove state pollution caused by test GsonFactoryTest.assertRegisterTypeAdapter to avoid order-dependent flaky tests

### LINKED ISSUE #1900: Flaky unit tests in GsonFactoryTest (non-idempotent)
### Expected behavior

Test `org.apache.shardingsphere.elasticjob.infra.json.GsonFactoryTest.assertRegisterTypeAdapter` should not pollute the state shared among tests, to prevent test dependency issues (i.e., a new test in the future could be flaky and fail when run after it due to the state pollution).

### Actual behavior

The test pollutes the state shared among tests. If run twice in the same JVM, the test would fail in the second run because the state pollution causes itself to fail.

The test pollutes some state shared among tests, causing the second test run to fail for assertions on the `GsonFactory`. The second run of `org.apache.shardingsphere.elasticjob.infra.json.GsonFactoryTest.assertRegisterTypeAdapter` fails for the following assertion:

```
Failed tests: 
  GsonFactoryTest.assertRegisterTypeAdapter:53 
Expected: is "{}"
     but: was "test"
```

Run the test twice in the same JVM. The second test run would fail. For example, one can make a duplicate of the method in the class like this:
```
@Test
public void assertRegisterTypeAdapterDuplicate() {
    assertRegisterTypeAdapter();
}
```
and then use maven to run the test class:
```
mvn test -pl elasticjob-infra/elasticjob-infra-common -am -Dtest=org.apache.shardingsphere.elasticjob.infra.json.GsonFactoryTest -DfailIfNoTests=false
```

y553546436:
Fixes #1900.

- Remove the `final` keyword for the `GsonBuilder` field in `GsonFactory` class.
- Add `clean` method in `GsonFactory` class, which re-initializes the `GsonBuilder`.
- Invoke the `clean` method in `GsonFactory` when the test `GsonFactoryTest.assertRegisterTypeAdapter` finishes.

With the proposed fix, the test does not pollute the shared state (and passes when run twice in the same JVM).

Note: this PR is a follow-up of #1774.

### DIFF
--- a/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/json/GsonFactory.java
+++ b/elasticjob-infra/elasticjob-infra-common/src/main/java/org/apache/shardingsphere/elasticjob/infra/json/GsonFactory.java
@@ -31,9 +31,9 @@ import lombok.NoArgsConstructor;
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class GsonFactory {
     
-    private static final GsonBuilder GSON_BUILDER = new GsonBuilder();
+    private static GsonBuilder gsonBuilder = new GsonBuilder();
     
-    private static volatile Gson gson = GSON_BUILDER.create();
+    private static volatile Gson gson = gsonBuilder.create();
     
     private static final JsonParser JSON_PARSER = new JsonParser();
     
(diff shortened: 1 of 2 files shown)


Technoboy-:
@y553546436 Nice catch.

Technoboy- (approved):

Technoboy-:
@TeslaCN

TeslaCN (approved):

Issue #1464, opened 2020-09-17
Suggest to support timeout for JobConfiguration


Technoboy-:
JobConfiguration should add timeout for users. If the job executes timeout event if, in sharding level, we can trace it and monitor it further.
As for dag, we can decide to wait for the max time for this node.

Technoboy-:
@terrymanu

terrymanu:
It is a necessary feature, I just set the label to new feature.

skaic:
I can make it, Could you assign to me. @terrymanu @TeslaCN

TeslaCN:
Hi @skaic 
We can discuss about how to support and what will happen if timed out.

skaic:
Thanks @TeslaCN .  It is a necessary feature for me ,  I have a idea and give me some time to write, Thanks.

skaic:
Core:
核心思想：

We start a process in new thread and waiting it。 When time out, call the interrupt() of the thread  And cloes() of the thread socket 。
我们通过线程启动任务后，使用一个计时器进行等待，如果超时了，就调用该线程的 interrupt 方法和该线程的套接字 close 方法。

So,  Ours Job can  try to find a timeout  By  Check  isInterrupted ( `Thread.currentThread().isInterrupted()` ) 、 catch InterruptedException | IOException , and to close self .
因此，我们的任务就可以通过检查 线程的 isInterrupted 标记(`Thread.currentThread().isInterrupted()`) 、 捕获 InterruptedException | IOException 来发现超时，让任务自己可以顺利的终止执行。




代码如下：

```
 for (int each : items) {
            ... 
            Future<Boolean> future = executorService.submit(() -> {
                try {
                    process(jobConfig, shardingContexts, each, jobExecutionEvent);
                } finally {
                    latch.countDown();
                }
                return true;
            });
            futures.add(future);
        }

        for (Future<Boolean> future : futures) {
            try {
                future.get(10, TimeUnit.SECONDS);
            } catch (final InterruptedException ex) {
                Thread.currentThread().interrupt();
            } catch (final ExecutionException ex) {
                throw new JobSystemException(ex);
            } catch (final TimeoutException ex) {
                // call the interrupt()
                future.cancel(true);
            }
        }
```

Job：


```
public class JavaSimpleJob implements SimpleJob {

    private final FooRepository fooRepository = FooRepositoryFactory.getFooRepository();

    @Override
    public void execute(final ShardingContext shardingContext) {
[...]

PR #1911, opened 2021-06-28
SingleSharding process keep pace with MultipleSharding process

### LINKED ISSUE #1910: ElasticJobExecutor execute have some different between  SingleSharding and MultipleSharding
## Question

look here：

```
@Slf4j
public final class ElasticJobExecutor {
  ......
  private void process(final JobConfiguration jobConfig, final ShardingContexts shardingContexts, final ExecutionSource executionSource) {
        Collection<Integer> items = shardingContexts.getShardingItemParameters().keySet();
        if (1 == items.size()) {
            int item = shardingContexts.getShardingItemParameters().keySet().iterator().next();
            JobExecutionEvent jobExecutionEvent = new JobExecutionEvent(IpUtils.getHostName(), IpUtils.getIp(), shardingContexts.getTaskId(), jobConfig.getJobName(), executionSource, item);
            process(jobConfig, shardingContexts, item, jobExecutionEvent);
            return;
        }
        CountDownLatch latch = new CountDownLatch(items.size());
        for (int each : items) {
            JobExecutionEvent jobExecutionEvent = new JobExecutionEvent(IpUtils.getHostName(), IpUtils.getIp(), shardingContexts.getTaskId(), jobConfig.getJobName(), executionSource, each);
            ExecutorService executorService = executorContext.get(ExecutorService.class);
            if (executorService.isShutdown()) {
                return;
            }
            executorService.submit(() -> {
                try {
                    process(jobConfig, shardingContexts, each, jobExecutionEvent);
                } finally {
                    latch.countDown();
                }
            });
        }
        try {
            latch.await();
        } catch (final InterruptedException ex) {
            Thread.currentThread().interrupt();
        }
    }
    ......
}
```

When `shardingContexts.getShardingItemParameters` size is 1 ,  process is run in currentThread。
Then  `shardingContexts.getShardingItemParameters` size not‘s 1， process is run in ExecutorService.Thread。

I  want to SingleSharding  process keep pace with  MultipleSharding process , what do you think?

skaic:
Fixes #1910.

- remove SingleSharding process run in currentThread
- catch ExecutionException and throw the JobSystemException of executorService.submit
-  `assertExecuteFailureWhenThrowExceptionForMultipleShardingItems`  test is JobSystemException

### DIFF
--- a/elasticjob-ecosystem/elasticjob-executor/elasticjob-executor-kernel/src/main/java/org/apache/shardingsphere/elasticjob/executor/ElasticJobExecutor.java
+++ b/elasticjob-ecosystem/elasticjob-executor/elasticjob-executor-kernel/src/main/java/org/apache/shardingsphere/elasticjob/executor/ElasticJobExecutor.java
@@ -135,31 +135,38 @@ public final class ElasticJobExecutor {
     
     private void process(final JobConfiguration jobConfig, final ShardingContexts shardingContexts, final ExecutionSource executionSource) {
         Collection<Integer> items = shardingContexts.getShardingItemParameters().keySet();
-        if (1 == items.size()) {
-            int item = shardingContexts.getShardingItemParameters().keySet().iterator().next();
-            JobExecutionEvent jobExecutionEvent = new JobExecutionEvent(IpUtils.getHostName(), IpUtils.getIp(), shardingContexts.getTaskId(), jobConfig.getJobName(), executionSource, item);
-            process(jobConfig, shardingContexts, item, jobExecutionEvent);
-            return;
-        }
         CountDownLatch latch = new CountDownLatch(items.size());
+        List<Future<Boolean>> futures = new ArrayList<>();
         for (int each : items) {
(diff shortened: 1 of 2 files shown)


TeslaCN (changes requested):
The Future and CountDownLatch are doing the same thing, aren't they?
My understanding is submitting a task to the thread pool may cause a thread context switch, which is unnecessary for a single shard.
I think it is better to keep it as it unless we have good reasons.

skaic:
Thanks for the reply. @TeslaCN 

I'm trying work of https://github.com/apache/shardingsphere-elasticjob/issues/1464 . 
So I need all process run are same `currentThread` or  `executorService.submit` at first.

TeslaCN:
> Thanks for the reply. @TeslaCN
> 
> I'm trying work of #1464 .
> So I need all process run are same `currentThread` or `executorService.submit` at first.

I think it is better to discuss about how before doing that.

skaic:
Yes, you are right.

PR #1922, opened 2021-07-06
Support timeout job

### LINKED ISSUE #1464: Suggest to support timeout for JobConfiguration

skaic:
Fixes #1464

- add timeout of JobConfiguration
- support timeout for ElasticJobExecutor


skaic:
close and repull by another account

PR #1956, opened 2021-08-06
bugfix: getJobConfiguration return null instead of throw NullPointerException

stanstanwang:
Bugfix: getJobConfiguration return null instead of throw NullPointerException

- check the zk result before `YamlEngine.unmarshal`

Before that, getting the JobConfiguration was more troublesome
``` java
JobConfigurationPOJO jobConfiguration = null;
try {
    jobConfiguration = jobConfigurationAPI.getJobConfiguration(jobKey);
} catch (NullPointerException e) {
    // ignore
}
// do something
```


TeslaCN (changes requested):
Hi @stanstanwang 
I think returning null is not elegant. How about throw something like `JobConfigurationException`?

elasticjob-lite/elasticjob-lite-lifecycle/src/main/java/org/apache/shardingsphere/elasticjob/lite/lifecycle/internal/settings/JobConfigurationAPIImpl.java
```diff
@@ -31,22 +31,26 @@
  */
 @RequiredArgsConstructor
 public final class JobConfigurationAPIImpl implements JobConfigurationAPI {
-    
+
```
TeslaCN: BTW, keep the indents please.

TeslaCN:
Code of conduct: https://shardingsphere.apache.org/community/en/contribute/code-conduct/

elasticjob-lite/elasticjob-lite-lifecycle/src/main/java/org/apache/shardingsphere/elasticjob/lite/lifecycle/internal/settings/JobConfigurationAPIImpl.java
stanstanwang: fixed，sorry but there is a litte strange why  4 space indent in every empty line

stanstanwang:
> Hi @stanstanwang
> I think returning null is not elegant. How about throw something like `JobConfigurationException`?

in my usage scenario , there is no check config exists method, so every time before i  get the config, i must catch the NullPointerException , this  is troublesome,  may be mark the method return @Nullable or using Optional is also elegant

Technoboy-:
> > Hi @stanstanwang
> > I think returning null is not elegant. How about throw something like `JobConfigurationException`?
> 
> in my usage scenario , there is no check config exists method, so every time before i get the config, i must catch the NullPointerException , this is troublesome, may be mark the method return @nullable or using Optional is also elegant

Currently, fit it that not throw NPE is enough. Other further discuss can open a new thread to discuss. @TeslaCN

TeslaCN:
Returning `null` is violate our code of conduct. I think `Optional` may be better.

Technoboy-:
> Returning `null` is violate our code of conduct. I think `Optional` may be better.

Where do this conflict?  If conflict, we should also refactor ZookeeperRegistryCenter. Not this API.

TeslaCN:
> > Returning `null` is violate our code of conduct. I think `Optional` may be better.
> 
> Where do this conflict? If conflict, we should also refactor ZookeeperRegistryCenter. Not this API.

This API is used by user directly.  Maybe the `ZookeeperRegistryCenter` does need to be refactored but not in this issue.

Technoboy-:
> > > Returning `null` is violate our code of conduct. I think `Optional` may be better.
> > 
> > 
> > Where do this conflict? If conflict, we should also refactor ZookeeperRegistryCenter. Not this API.
> 
> This API is used by user directly. Maybe the `ZookeeperRegistryCenter` does need to be refactored but not in this issue.

If user upgrade to current version, compile error in they project.

elasticjob-lite/elasticjob-lite-lifecycle/src/main/java/org/apache/shardingsphere/elasticjob/lite/lifecycle/internal/settings/JobConfigurationAPIImpl.java
```diff
@@ -36,7 +36,11 @@
     
     @Override
     public JobConfigurationPOJO getJobConfiguration(final String jobName) {
-        return YamlEngine.unmarshal(regCenter.get(new JobNodePath(jobName).getConfigNodePath()), JobConfigurationPOJO.class);
+        String yamlContent = regCenter.get(new JobNodePath(jobName).getConfigNodePath());
+        if (yamlContent == null) {
```
TeslaCN: Constant is on the left.

TeslaCN:
> > > > Returning `null` is violate our code of conduct. I think `Optional` may be better.
> > > 
> > > 
> > > Where do this conflict? If conflict, we should also refactor ZookeeperRegistryCenter. Not this API.
> > 
> > 
> > This API is used by user directly. Maybe the `ZookeeperRegistryCenter` does need to be refactored but not in this issue.
> 
> If user upgrade to current version, compile error in they project.

OK, we can keep the current API. But we should polish it in future.

elasticjob-lite/elasticjob-lite-lifecycle/src/main/java/org/apache/shardingsphere/elasticjob/lite/lifecycle/internal/settings/JobConfigurationAPIImpl.java
stanstanwang: fixed, sorry for that..

elasticjob-lite/elasticjob-lite-lifecycle/src/main/java/org/apache/shardingsphere/elasticjob/lite/lifecycle/internal/settings/JobConfigurationAPIImpl.java
```diff
@@ -36,7 +36,11 @@
     
     @Override
     public JobConfigurationPOJO getJobConfiguration(final String jobName) {
-        return YamlEngine.unmarshal(regCenter.get(new JobNodePath(jobName).getConfigNodePath()), JobConfigurationPOJO.class);
+        String yamlContent = regCenter.get(new JobNodePath(jobName).getConfigNodePath());
```
Technoboy-, line 39: How about like this:
Optional.ofNullable(yamlContent).map(v -> YamlEngine.unmarshal(v, JobConfigurationPOJO.class)).orElse(null) ?
@TeslaCN @stanstanwang

Technoboy- (approved):
LGTM

TeslaCN (approved):

Issue #1958, opened 2021-08-12
some tasks run very frequently and do not want to log. I want to configure logging separately for tasks with long running intervals


Fdemo123:
some tasks run very frequently and do not want to log. I want to configure logging separately for tasks with long running intervals

sunkai-cai:
For English only.

Are you using `elasticjob-lite-spring-boot-starter` package?  

This a good idea, maybe we can support it in next version.

TeslaCN:
Please translate your issue into English.

Fdemo123:
> For English only.
> 
> Are you using `elasticjob-lite-spring-boot-starter` package?
> 
> This a good idea, maybe we can support it in next version.

I am using elastic job version 2.1.5. I have written a springboot starter to handle the loading of tasks and control whether each task records logs separately. I see that the official website is updated. I want to upgrade and find that this function is not available. I hope the official can support it. In this way, the developer does not need to do secondary encapsulation.

sunkai-cai:
> > For English only.
> > Are you using `elasticjob-lite-spring-boot-starter` package?
> > This a good idea, maybe we can support it in next version.
> 
> I am using elastic job version 2.1.5. I have written a springboot starter to handle the loading of tasks and control whether each task records logs separately. I see that the official website is updated. I want to upgrade and find that this function is not available. I hope the official can support it. In this way, the developer does not need to do secondary encapsulation.

Of course,  the features in the plan, and welcome the developer submit PR.  Could you contribute code to complete the plan faster .

TeslaCN:
Event trace is not a global configuration. You can configure event trace for each job separately.

PR #1963, opened 2021-08-17
enableEventTrace Separate configuration

Fdemo123:
Fixes #ISSUSE_ID.
https://github.com/apache/shardingsphere-elasticjob/issues/1958

- Added enableEventTrace attribute in ElasticJobConfigurationProperties,JobConfigurationPOJO,JobConfiguration,JobExecutionEvent
- The modified method may cause some inconsistencies between the implementation and naming of the postJobExecutionEvent and postJobStatusTraceEvent methods in the JobFacade, but this should be done with minimal changes
- Adding the enableeventtrace attribute will cause an error on the task list page in elasticjob-3.0.0-lite-ui:
Unable to find property 'enableEventTrace' on class: org.apache.shardingsphere.elasticjob.infra.pojo.JobConfigurationPOJO

### DIFF
--- a/elasticjob-api/src/main/java/org/apache/shardingsphere/elasticjob/api/JobConfiguration.java
+++ b/elasticjob-api/src/main/java/org/apache/shardingsphere/elasticjob/api/JobConfiguration.java
@@ -80,6 +80,8 @@ public final class JobConfiguration {
     private final String label;
     
     private final boolean staticSharding;
+
+    private final boolean enableEventTrace;
     
     /**
      * Create ElasticJob configuration builder.
@@ -138,6 +140,8 @@ public final class JobConfiguration {
         private String label;
         
         private boolean staticSharding;
+
+        private boolean enableEventTrace;
     
         /**
          * Cron expression.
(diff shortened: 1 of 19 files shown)


TeslaCN:
@Fdemo123 
Event trace is not a global configuration. You can configure event trace for each job separately.

TeslaCN:
@Fdemo123 I'm sorry that this PR is not right. I'm closing it. It's better to discuss before submitting a PR.

Fdemo123:
> @Fdemo123
> Event trace is not a global configuration. You can configure event trace for each job separately.

 Event trace It's not really a global configuration,I think event trace and cron belong to the same level,so the configuration is as follows:
    simpleJob:
      elasticJobClass: org.apache.shardingsphere.elasticjob.lite.example.job.SpringBootSimpleJob
      cron: 0/5 * * * * ?
      shardingTotalCount: 3
      shardingItemParameters: 0=Beijing,1=Shanghai,2=Guangzhou
      enableEventTrace: true

PR #2021, opened 2021-12-17
Fix itemErrorMessages not cleared after job finished

### LINKED ISSUE #2019: ElasticJobExecutor.itemErrorMessages never cleared?
I expend a custom tracing configuration to record JobExecutionEvent, and JobTraceStatusEvent.
and I met a problem.

The job execute with an Runtime Exception at first time.
So, I received a record with state JobStatusTraceEvent.State.TASK_ERROR, and some detail error message.
It's worked expectly.

And then, the job execute again successfully without any exception or error.
But, I still received a record with state JobStatusTraceEvent.State.TASK_ERROR.
And, the state and message is as same as first time run.
I think, this is unexpectly.

Is it a bug or if I misunderstood it?

After that, I read the source code In file org.apache.shardingsphere.elasticjob.executorElasticJobExecutor.java
I found that the ElasticJobExecutor.itemErrorMessages was initialed when job registed, and put message when catch a Throwable,
But, the map is never cleared.

ilyng:
It will make error message show again in next loop.

Fixes #2019.

-
-
-


TeslaCN (approved):
@smartBun Thank you!

PR #2027, opened 2021-12-28
Add unit case for JobSystemException

cocodroid:
Add unit case for JobSystemException


TeslaCN (approved):
@cocodroid Thank you!

PR #2030, opened 2021-12-29
[Refactor] Add enum related comment

springmonster:
Add enum related comment


TeslaCN (changes requested):
@springmonster Don't you think the naming in codes are clear enough to understand?

springmonster:
@TeslaCN , yes, you're right. The existed code is self-explanatory. However, enum is somewhat inconsistent in comments, and lint will issue a warning.
Should I exclude enum missing comments warning from my side and ignore this situation?
Thanks very much!

TeslaCN:
> @TeslaCN , yes, you're right. The existed code is self-explanatory. However, enum is somewhat inconsistent in comments, and lint will issue a warning. Should I exclude enum missing comments warning from my side and ignore this situation? Thanks very much!

@springmonster I think the comments added in this PR are redundant. You may ignore the lint in your IDE. But thanks for your contributions still.

PR #2087, opened 2022-05-19
Merge similar GitHub Actions workflow files #2083

### LINKED ISSUE #2083: Merge similar GitHub Actions workflow files
## Refactor

Workflow files are highly similar, which can be refactored by matrix and merged into one workflow.
https://github.com/apache/shardingsphere-elasticjob/tree/master/.github/workflows

m1ngyuan:
Fixes #2083.

- Merge similar GitHub Actions workflow files


pom.xml
```diff
@@ -705,6 +705,12 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
```
TeslaCN: What's the different with `-Dmaven.javadoc.skip=true`?

TeslaCN (changes requested):

pom.xml
m1ngyuan: -Dmaven.javadoc.skip=true on windows will be fail

TeslaCN: > -Dmaven.javadoc.skip=true on windows will be fail

Why? Could you check it again? I think it doesn't make sense.

m1ngyuan: Thank you for your advice,I found a solution to the problem,Please trigger GitHub workflow again

TeslaCN (approved):
Thank you!

PR #2092, opened 2022-06-10
Each job has its own listener event notify thread instead of the only one curator thread

### LINKED ISSUE #2038: OneOffJob cannot be triggered or can only be triggerd on a part of instances occasionally
Version:  ElasticJob-Lite 3.0.1 & Curator 5.2.0
    Problem: All instances that have sharding items should start running after calling instanceService.triggerAllInstances(), but some instances cannot be triggered.
    The problem occured occasionally, and it disappear if we restart the instance or any other instance join in, which will cause the resharding nessary flag being setted.
The dump when the problem occured:
任务触发后卡住时注册中心数据结构快照(网贴用)
    From the dump we can see that  trigger node have instances nodes that didn't be triggered, the instances ip1@-8600 and ip2@-@3732 are running normally but their shards cannot be triggered. 
    Any relay will be appreaciated!

RedRed228:
Fixes #2038.

- Add ListenerNotifierManager to manage job listener's notify thread executor. The thread executor is created for the job when 
ListenerManager is constructed.
- Change CoordinatorRegistryCenter's watch method:
  old: watch(String key, DataChangedEventListener listener) ;
  new: watch(String key, DataChangedEventListener listener, Executor executor);

Each job now has its own event notify thread, thus no influence would occur between different jobs.

### DIFF
--- a/elasticjob-infra/elasticjob-registry-center/elasticjob-regitry-center-provider/elasticjob-registry-center-zookeeper-curator/src/main/java/org/apache/shardingsphere/elasticjob/reg/zookeeper/ZookeeperRegistryCenter.java
+++ b/elasticjob-infra/elasticjob-registry-center/elasticjob-regitry-center-provider/elasticjob-registry-center-zookeeper-curator/src/main/java/org/apache/shardingsphere/elasticjob/reg/zookeeper/ZookeeperRegistryCenter.java
@@ -58,6 +58,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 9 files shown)


elasticjob-infra/elasticjob-registry-center/elasticjob-regitry-center-provider/elasticjob-registry-center-zookeeper-curator/src/main/java/org/apache/shardingsphere/elasticjob/reg/zookeeper/ZookeeperRegistryCenter.java
```diff
@@ -408,20 +409,28 @@ public void executeInLeader(final String key, final LeaderExecutionCallback call
     }
     
     @Override
-    public void watch(final String key, final DataChangedEventListener listener) {
+    public void watch(final String key, final DataChangedEventListener listener, final Executor executor) {
         CuratorCache cache = caches.get(key + "/");
-        cache.listenable().addListener((curatorType, oldData, newData) -> {
-            if (null == newData && null == oldData) {
-                return;
+        CuratorCacheListener cacheListener = new CuratorCacheListener() {
```
TeslaCN: Why change this to anonymous class?

RedRed228: > Why change this to anonymous class?

I'm not familiar with lamda expression,  so I use anonymous class to resuse the code. I can change it back to lamda expression.

TeslaCN (approved):

PR #2123, opened 2022-09-01
fix leader nullpointexception and increament listener core pool size

### LINKED ISSUE #2122: Frequently creation and shutdown cause zookeeper watch to be lost.
master

ElasticJob-Lite

### Expected behavior
every event in zookeeper, client can watch it, otherwise exception occur

### Actual behavior
frequently create and shutdown Immediately，watch lost

project has 12 listener now, every one can watch event, but single thread pool core size is not enough, when job shutdown, shudownListener maybe execute over first, but GuaranteeListener maybe not get complete signal，lead to AbstractDistributeOnceElasticJobListener appear timeout exception，so should be increament core size to avoid watch loss.

extends a AbstractDistributeOnceElasticJobListener,  set complete or start timeout, Frequently creation and shutdown some job

### Example codes for reproduce this issue (such as a github link).

xlcn:
Fixes #2122 . 

-
-
-
when job shutdown,  code should check before elect leader.
increament listener threadpool core size , avoid watch lost

### DIFF
--- a/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/listener/ListenerNotifierManager.java
+++ b/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/listener/ListenerNotifierManager.java
@@ -63,8 +63,8 @@ public final class ListenerNotifierManager {
             synchronized (this) {
                 if (!listenerNotifyExecutors.containsKey(jobName)) {
                     ThreadFactory threadFactory = ThreadUtils.newGenericThreadFactory("ListenerNotify-" + jobName);
-                    // The thread pool count should be as consistent as possible with the listener count, avoid watch loss.
-                    ExecutorService notifyExecutor = Executors.newFixedThreadPool(15, threadFactory);
+                    // Use unbounded thread pool to cover all the listeners to execute event, avoid zk watch loss.
+                    ExecutorService notifyExecutor = Executors.newCachedThreadPool(threadFactory);
                     listenerNotifyExecutors.put(jobName, notifyExecutor);
                 }
             }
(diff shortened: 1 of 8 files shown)


TeslaCN:
How about using unbounded thread pool?

xlcn:
Thank you for your advice. please recheck.

elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/listener/ListenerNotifierManager.java
```diff
@@ -63,7 +63,8 @@ public void registerJobNotifyExecutor(final String jobName) {
             synchronized (this) {
                 if (!listenerNotifyExecutors.containsKey(jobName)) {
                     ThreadFactory threadFactory = ThreadUtils.newGenericThreadFactory("ListenerNotify-" + jobName);
-                    ExecutorService notifyExecutor = Executors.newSingleThreadExecutor(threadFactory);
+                    // Use unbounded thread pool to cover all the listeners to execute event, avoid zk watch loss.
```
TeslaCN: The comment could be removed.

TeslaCN:
Please fix the CI.

elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/listener/ListenerNotifierManager.java
xlcn: okay

xlcn:
retry? local machine (mac) test is okay

TeslaCN:
> retry? local machine (mac) test is okay

That means the unit tests are not stable.

xlcn:
> > retry? local machine (mac) test is okay
> 
> That means the unit tests are not stable.

timeout again _^!

xlcn:
@TeslaCN could you rerun CI action? Thx

PR #2184, opened 2023-02-21
make startup faster

### LINKED ISSUE #2185: make startup  faster
#  Is your feature request related to a  #problem?

I think the same dataSource just need one RDBJobEventStorage instance 。after all, init RDBJobEventStorage  will take some time to get a rdb connection  and check whether the table and index exist

The benefit of the change is that the startup will be  #faster

In the experiment, starting 100 jobs can save about 8 seconds, of course, this depends on the environment, if the database and application are on different networks may be slower
#2184

wizhuo:
Fixes #2185 

I think the same dataSource just need one RDBJobEventStorage instance 。after all, init RDBJobEventStorage  will take some time to get a rdb connection  and check whether the table and index exist

The benefit of the change is that the startup will be  #faster

In the experiment, starting 100 jobs can save about 8 seconds, of course, this depends on the environment, if the database and application are on different networks may be slower


elasticjob-ecosystem/elasticjob-tracing/elasticjob-tracing-rdb/src/main/java/org/apache/shardingsphere/elasticjob/tracing/rdb/storage/RDBJobEventStorage.java
```diff
@@ -61,14 +63,41 @@ public final class RDBJobEventStorage {
...
+        return wrapException(() -> STORAGE_MAP.computeIfAbsent(dataSource, (ds) -> {
+            try {
+                return new RDBJobEventStorage(ds);
+            } catch (SQLException e) {
+                throw new RuntimeException(e);
+            }
+        }));
+    }
+
+    public static RDBJobEventStorage wrapException(Callable<RDBJobEventStorage> callable) throws SQLException {
```
TeslaCN: How about using `Supplier` instead of `Callable`?

wizhuo: I would love to! I've modified it, please check it out

wizhuo: image
https://github.com/apache/shardingsphere-elasticjob/actions/runs/4248881615/jobs/7388688271
 
The above link is the information about the failure of the last build. From the error information, we can see that zookeeper was closed for some reason, which caused the unit test to fail. I tried executing the command locally with success: 

```
./mvnw --batch-mode --no-transfer-progress '-Dmaven.javadoc.skip=true' clean install
```
I  can not reproduce the above error, what do you suggest?

elasticjob-ecosystem/elasticjob-tracing/elasticjob-tracing-api/src/main/java/org/apache/shardingsphere/elasticjob/tracing/exception/WrapException.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * use to wrap Exception.
+ * @author willJo
+ */
+
+public class WrapException extends RuntimeException {
+    public WrapException(final Throwable cause) {
+        super(cause);
+    }
```
TeslaCN:
```suggestion
/**
 * Use to wrap Exception.
 */
public class WrapException extends RuntimeException {
    
    public WrapException(final Throwable cause) {
        super(cause);
    }
```

`@author` is not applicable here.

elasticjob-ecosystem/elasticjob-tracing/elasticjob-tracing-rdb/src/main/java/org/apache/shardingsphere/elasticjob/tracing/rdb/storage/RDBJobEventStorage.java
```diff
@@ -55,26 +58,63 @@ public final class RDBJobEventStorage {
...
+            try {
+                return new RDBJobEventStorage(ds);
+            } catch (SQLException e) {
+                throw new WrapException(e);
+            }
+        }));
+    }
+
+    /**
+     * wrapException util method.
```
TeslaCN:
```suggestion
     * WrapException util method.
```

elasticjob-ecosystem/elasticjob-tracing/elasticjob-tracing-rdb/src/main/java/org/apache/shardingsphere/elasticjob/tracing/rdb/storage/RDBJobEventStorage.java
```diff
@@ -55,26 +58,63 @@ public final class RDBJobEventStorage {
...
+    private RDBJobEventStorage(final DataSource dataSource) throws SQLException {
         this.dataSource = dataSource;
         databaseType = getDatabaseType(dataSource);
         sqlMapper = new RDBStorageSQLMapper(databaseType.getSQLPropertiesFile());
         initTablesAndIndexes();
     }
-    
+
+    /**
+     * the same dataSource  always return the same RDBJobEventStorage instance.
```
TeslaCN:
```suggestion
     * The same dataSource always return the same RDBJobEventStorage instance.
```

TeslaCN (approved):

PR #2211, opened 2023-04-17
add zk listener remove to avoid memory leak.

### LINKED ISSUE #2210: Memory leak when the task shutdown.
3.0.2

ElasticJob-Lite

### Expected behavior
when the task shutdown, all instances of the task could be GC later 

### Actual behavior
watch listener of the task hold some instance, has gc root, can not gc
image

when the task shutdown, watch listener has not remove.

shutdown a task, check the left instance of heap dump

xlcn:
Fixes #2210 

- add zk listener remove to avoid memory leak.
-
-


elasticjob-infra/elasticjob-registry-center/elasticjob-regitry-center-provider/elasticjob-registry-center-zookeeper-curator/src/main/java/org/apache/shardingsphere/elasticjob/reg/zookeeper/ZookeeperRegistryCenter.java
```diff
@@ -71,7 +74,19 @@ public final class ZookeeperRegistryCenter implements CoordinatorRegistryCenter
     private final ZookeeperConfiguration zkConfig;
     
     private final Map<String, CuratorCache> caches = new ConcurrentHashMap<>();
-    
+
+    /**
+     * Data listener list.
+     */
+    @Getter(AccessLevel.PROTECTED)
```
TeslaCN: It seems that the Getter is used by test code only. Could we consider removing this and get field by Mockito Plugins or reflection in test code.

TeslaCN (approved):
Thank you.

PR #2251, opened 2023-09-01
remove offline server ip

### LINKED ISSUE #2229: Change the ip node under servers to a temporary node
After reading the source code of the job, I have a question, why not set the ip child node under the servers node as a temporary node, but save it permanently, so that with the release of services in k8s, there will be more and more ip child nodes , and these ip nodes will be traversed when judging the job status, which will cause problems as the number increases.
I think it is feasible to set the node under servers as a temporary node. After the release, the old ip is useless, and it doesn't matter if it is deleted. I will do a test locally.
Regarding this point, I may have some inconsiderations, and I hope the official can see if it is feasible.

`    /**
     * 持久化作业服务器上线信息.
     * 
     * @param enabled 作业是否启用
     */
    public void persistOnline(final boolean enabled) {
        if (!JobRegistry.getInstance().isShutdown(jobName)) {
//            jobNodeStorage.fillJobNode(serverNode.getServerNode(JobRegistry.getInstance().getJobInstance(jobName).getIp()), enabled ? "" : ServerStatus.DISABLED.name());
            //改为临时节点
            jobNodeStorage.fillEphemeralJobNode(serverNode.getServerNode(JobRegistry.getInstance().getJobInstance(jobName).getIp()), enabled ? "" : ServerStatus.DISABLED.name());
        }
    }
    `

### LINKED ISSUE #2239: 怎么清理垃圾数据
elastic-job运行会在zk里面写数据，我们是容器服务，每次发布重启就会向zookeeper的servers里面写入ip信息，这样历史数据越积越多，这个elastic-job有提供好的办法吗？

### LINKED ISSUE #2241: how to delete data in /servers
our application in kubernets ,when udate the image the pod ip will be change .it will result many history ip in servers ,how should i delete it

songxiaosheng:
Fixes #ISSUSE_ID.
- fix #2239
- fix #2229
- fix #2241

- support remove offline serverIp logic

### DIFF
--- a/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/server/ServerService.java
+++ b/elasticjob-lite/elasticjob-lite-core/src/main/java/org/apache/shardingsphere/elasticjob/lite/internal/server/ServerService.java
@@ -18,6 +18,7 @@
 package org.apache.shardingsphere.elasticjob.lite.internal.server;
 
 import com.google.common.base.Strings;
+import org.apache.commons.lang.StringUtils;
 import org.apache.shardingsphere.elasticjob.lite.internal.instance.InstanceNode;
 import org.apache.shardingsphere.elasticjob.lite.internal.schedule.JobRegistry;
 import org.apache.shardingsphere.elasticjob.lite.internal.storage.JobNodeStorage;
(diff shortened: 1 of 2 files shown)


songxiaosheng:
@TeslaCN  PATL

TeslaCN:
Could we add an option for this? Sometimes we may set server disable in non-container environment.

songxiaosheng:
> Could we add an option for this? Sometimes we may set server disable in non-container environment.

This is compatible, no stateful nodes are removed, and the existing state does not remove the recommended hosting state.

TeslaCN (approved):

TeslaCN (approved):

PR #2305, opened 2023-10-20
fix status badge URLs

shoito:
See https://github.com/badges/shields/issues/8671

> You need to update your badge URL from
> https://img.shields.io/github/workflow/status/<user>/<repo>/Run%20Tests
> to
> https://img.shields.io/github/actions/workflow/status/<user>/<repo>/test.yml?branch=main


linghengqian (approved):
- Thanks for the PR. This badge is very embarrassing because the use of Travis CI should be removed at the request of the Apache Infra team. Refer to https://infra.apache.org/ .

> February 3, 2023: The Infrastructure team has been working on decommissioning Travis CI support since late 2022, due to changes in the Travis Open Source policy. We are now 10 days prior to the announced cut off date of February 15, and some projects still have one or more builds configured to use Travis CI. While we understand that not all projects may be finished migrating their builds to a supported platform, we will be doing the cut-off on the 15th. Any PMC with concerns regarding its projects should state them in an Infra Jira ticket so that we can track build migration and provide assistance / direction to folks who request it.

- Travis CI has been removed on the side of the parent project ShardingSphere, and perhaps in the future, the project PMC will have time to push this forward. But this does not prevent the current PR from being merged to improve the README experience.

PR #2438, opened 2024-09-24
optimize TracingProperties

ooooooook:
- optimize TracingProperties, add spring-configuration-metadata for nested DataSourceProperties

### DIFF
--- a/spring/boot-starter/src/main/java/org/apache/shardingsphere/elasticjob/spring/boot/tracing/ElasticJobTracingConfiguration.java
+++ b/spring/boot-starter/src/main/java/org/apache/shardingsphere/elasticjob/spring/boot/tracing/ElasticJobTracingConfiguration.java
@@ -21,7 +21,6 @@ import com.zaxxer.hikari.HikariDataSource;
 import org.apache.shardingsphere.elasticjob.kernel.tracing.config.TracingConfiguration;
 import org.springframework.beans.BeanUtils;
 import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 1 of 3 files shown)


linghengqian (changes requested):
- The PR did not pass the code formatting validation. You can execute first `./mvnw spotless:apply -Pcheck -T1C`. Then execute `./mvnw checkstyle:check -Pcheck -T1C` to manually adjust to repair the CI.

spring/boot-starter/src/main/java/org/apache/shardingsphere/elasticjob/spring/boot/tracing/ElasticJobTracingConfiguration.java
```diff
@@ -68,9 +67,11 @@ public DataSource tracingDataSource(final TracingProperties tracingProperties) {
          * @return a bean of tracing configuration
          */
         @Bean
-        @ConditionalOnBean(DataSource.class)
-        public TracingConfiguration<DataSource> tracingConfiguration(@Qualifier("dataSource") final DataSource dataSource,
+        public TracingConfiguration<DataSource> tracingConfiguration(@Qualifier("dataSource") @Nullable final DataSource dataSource,
                                                                      @Qualifier("tracingDataSource") @Nullable final DataSource tracingDataSource) {
+            if (dataSource == null && tracingDataSource == null) {
+                throw new IllegalArgumentException("Neither dataSource nor tracingDataSource bean is required");
+            }
```
linghengqian, line 74: - I would say this is a bit illogical. In fact, `org.apache.shardingsphere.elasticjob:elasticjob-spring-boot-starter` should not have a mandatory dependency on `org.springframework.boot:spring-boot-starter-jdbc`.
- I may raise another PR later to try to remove the mandatory dependency on `org.springframework.boot:spring-boot-starter-jdbc`. Or are you going to do that? Because the Spring Bean for `TracingConfiguration` should actually only be registered when `elasticjob.tracing.type` is `RDB`.

ooooooook: The dependency `org.springframework.boot:spring-boot-starter-jdbc` is optional. It attempts to retrieve the DataSource from `elasticjob.tracing.dataSource`. If it is not found, it will then obtain the default DataSource from context.  @linghengqian

linghengqian: - I missed the context of the current class earlier. The current PR looks fine. I need a little time to see why CI is failing with unstable.

linghengqian (approved):
- LGTM!
- The instability of CI is caused externally and does not need to be discussed in the current PR.

Issue #2413, opened 2024-07-30
Integrate develocity.apache.org in the project to prepare for possible performance testing


linghengqian:
## Feature Request

**For English only**, other languages will not accept.

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot make decision by current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Is your feature request related to a problem?

- See https://github.com/apache/shardingsphere-elasticjob/issues/2258#issuecomment-1721423104 .

> `@zjx990`:
> When conducting this kind of automated testing, can we introduce a scenario? Specifically, consider a client or, more accurately, a server that actually executes jobs, reaching its limits. 
> For example: This instance is an execution node for 900 jobs, which is the actual server. These 900 jobs are scheduled every second. When would this server reach its capacity limit?
> My thought: My idea is to focus on its CPU information, since job scheduling is periodic, and thus the CPU usage will also fluctuate periodically. I'll ignore the memory used by the jobs for now. If the CPU peak reaches 70%, then I would consider the limit of this instance to be 900 jobs.

### Describe the feature you would like.

- Integrate ge.apache.org in the project to prepare for possible performance testing. Refer to https://infra.apache.org/blog/gradle-has-landed-at-the-asf.html .
- It is not intuitive to perform performance testing on ShardingSphere ElasticJob in CI, which is limited by the fact that Github Actions only provides logs and does not provide much CI data visualization. Using https://ge.apache.org/ to improve the observability of CI may be a good way forward. This should help compare performance changes before and after the project changes.

linghengqian:
- ge.apache.org is currently using Develocity 2024.1.7, which looks more likely to be unusable than I thought.
- ![image](https://github.com/user-attachments/assets/d4388039-efa1-459e-b474-5789c80d89b9)

linghengqian:
This still cannot get around the limitation that Github Actions only has 4 cores and 16gb of memory.

linghengqian:
It seems to be more stable after migrating to https://develocity.apache.org/ . See https://infra.apache.org/gradle.html .

linghengqian:
I sent an email to `gradle-enterprise-infrastructure-team@gradle.com` yesterday to confirm some experience improvements for https://develocity.apache.org/ . Currently, https://develocity.apache.org/ lacks i18n support and an Android app.

PR #2441, opened 2024-09-29
add destroyMethod for zookeeperRegistryCenter bean

ooooooook:
Fixes #ISSUSE_ID.

- Add a `destroyMethod` for the `zookeeperRegistryCenter` bean to ensure that the application removes the Zookeeper ephemeral node upon exit.


linghengqian (approved):
- LGTM
- I still can't determine why the CI failed so quickly. In contrast, the CI for #2442 did not suffer from this. Proposal #2413 is worth further consideration, but with https://ge.apache.org/ frequently going offline, I'm not sure if Gradle Inc. or ASF Infra has a bigger problem.

PR #2445, opened 2024-10-08
fix remove offline server

ooooooook:
Fixes #ISSUSE_ID.

- fix remove offline server.  (Unable to remove offline server on server start up)

### DIFF
--- a/kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/server/ServerService.java
+++ b/kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/server/ServerService.java
@@ -109,7 +109,7 @@ public final class ServerService {
     
     /**
      *  Remove unused server IP.
-     *  
+     *
      * @return num of server IP to be removed
      */
     public int removeOfflineServers() {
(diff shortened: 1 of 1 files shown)


linghengqian (commented):
- LGTM
- @songxiaosheng Is this an omission from #2251?

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/server/ServerService.java
```diff
@@ -132,7 +132,7 @@ public int removeOfflineServers() {
                 return;
             }
             String status = jobNodeStorage.getJobNodeData(serverNode.getServerNode(serverIp));
-            if (StringUtils.isNotBlank(status)) {
+            if (StringUtils.isBlank(status)) {
```
linghengqian, line 135: - Is this actually corresponding to `org.apache.shardingsphere.elasticjob.kernel.internal.server.ServerStatus`?
```java
public enum ServerStatus {
    ENABLED, DISABLED
}
```
- Then why not just `return` when `ENABLED` is determined?

ooooooook: - Only offline servers return null, online servers can return enabled or disabled
- We should not remove the server while it is still online, even if the job is disabled.
@linghengqian

PR #2462, opened 2024-12-08
feature: https://github.com/apache/shardingsphere-elasticjob/issues/2461

### LINKED ISSUE #2461: Support New Sharding Strategy To Run Balanceful On Multiple Servers For Single Sharding Scenario
## Feature Request

**For English only**, other languages will not accept.

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot make decision by current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

Yes, here is the detail information.

We use elastic job as job framework in our job applications heavily, however our job applications are always running a job in single sharding which means the sharding total count is only 1, and our job applications are always deploying on more than two servers.

You know, there are many job tasks in one job application, some job tasks maybe run quickly and need a little resources, but some job tasks are very heavy, need much of resources and running will be elapsed a lot time. The problem is for a certain job task is always running on a certain server after sharding scheduled, that caused one server would occupy much more CPU and memory for all the heavy tasks are running on, but the other server is more idle for the running on tasks are lightly.

we want the resources like cpu, memory will be used balanced, and not waste some resources just for a standby failover.

We want one job task to run balanced in multiple servers. 

For example, there is a heavy task that reading file and compare with the DB records then do business handling, and there are 3 servers A, B, C  used to deploy on. Currently, the builtin sharding strategy in the framework can only support to split all the jobs and assign them to different server, like using 'ROUND_ROBIN' job sharding strategy type, but the assigned job is sticked with the server, let's say it is assigned to server A, then can't be changed.

However, what we want to achieve is this heavy task can run on multiple servers balanced, like it is firstly running on server A, then the second time it will be running on server B, then the third time it will be running on server C, and the next time it will be running on server A again, and like this in turn.

### Solution

The elastic job version what we currently using is 3.0.1, and I already resolve this problem and finish the solution of this feature. I deploy our job applications that using this new feature in our product environment, it works well.

I have searched all the open issues that there is no issue talked about this feature, and also I check the codes above version 3.0.1, eventually the latest master codes, no feature about this. 

So I post this ticket and want to discuss here, could this feature be supported in the framework? if it is accepted, I could contribute  the codes that I have written to the community.

baiyangzhuhong:
Fixes #2461.

## Intent

As #2461  figured out, this post is to enhancement job strategy balance in the case of single sharding. 
We want to let the job run on every job instance balanced like 'Round Robin' behavior, but not stick with only one certain instance.

## Solution

Basically, the design is add new single sharding handling logic which can decouple with exists logic, and can be good benefits on expanding in the future.

So there is new `SingleShardingBalanceJobShardingStrategy` and added in the SPI `/META-INF/services/org.apache.shardingsphere.elasticjob.kernel.internal.sharding.strategy.JobShardingStrategy` list. On the other side, I do codes refactor on the job facade for the entrance of job sharding strategy action. So the `JobFacade` class is modified to an interface, and we have a new `AbstractJobFacade` abstract class, then decouple the default sharding facade with single sharding facade. I do believe making `JobFacade` as an interface is better, this is also a good design following 'Open-close' principle.

```text
                                              JobFacade
                                                 |
                                                \|/
                                           AbstractJobFacade
                                                 |
                                                / \
                                            /        \
                                         /              \
                                ShardingJobFacade     SingleShardingJobFacade
```

### SingleShardingBalanceJobShardingStrategy

Obviously, if we want to do  job running balanced in distribution multiple job instance environment, we need a center controlling on the balancement status, but I don't want this behavior to be heavy. So I use ephemeral node handling in zookeeper, this can be promised that one job running on every job instance balanced is guarded in normal running. And for the other scenarios, you know, some unexpected cases, is totally fallback to the default logic. It is very lightly and effiecent.

I also add unit tests on the new classes, and modified the exists tests code.

Please do check. 

Thanks

### DIFF
--- a/kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
+++ b/kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
@@ -82,14 +82,15 @@ public final class JobScheduler {
         Collection<ElasticJobListener> jobListeners = getElasticJobListeners(this.jobConfig);
         setUpFacade = new SetUpFacade(regCenter, this.jobConfig.getJobName(), jobListeners);
         schedulerFacade = new SchedulerFacade(regCenter, this.jobConfig.getJobName());
-
-        if (1 == this.jobConfig.getShardingTotalCount() // the single sharding scenario
-            && "SINGLE_SHARDING_BALANCE".equals(this.jobConfig.getJobShardingStrategyType())) { // the specified SINGLE_SHARDING_BALANCE strategy
+        
+        // the single sharding scenario and specified SINGLE_SHARDING_BALANCE strategy
+        if (1 == this.jobConfig.getShardingTotalCount()
+                && "SINGLE_SHARDING_BALANCE".equals(this.jobConfig.getJobShardingStrategyType())) {
             jobFacade = new SingleShardingJobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
         } else {
             jobFacade = new ShardingJobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
         }
-
+        
         validateJobProperties();
         jobExecutor = new ElasticJobExecutor(elasticJob, this.jobConfig, jobFacade);
         setGuaranteeServiceForElasticJobListeners(regCenter, jobListeners);
@@ -103,14 +104,15 @@ public final class JobScheduler {
         Collection<ElasticJobListener> jobListeners = getElasticJobListeners(this.jobConfig);
         setUpFacade = new SetUpFacade(regCenter, this.jobConfig.getJobName(), jobListeners);
         schedulerFacade = new SchedulerFacade(regCenter, this.jobConfig.getJobName());
-
-        if (1 == this.jobConfig.getShardingTotalCount() // the single sharding scenario
-            && "SINGLE_SHARDING_BALANCE".equals(this.jobConfig.getJobShardingStrategyType())) { // the specified SINGLE_SHARDING_BALANCE strategy
+        
+        // the single sharding scenario and specified SINGLE_SHARDING_BALANCE strategy
+        if (1 == this.jobConfig.getShardingTotalCount()
+                && "SINGLE_SHARDING_BALANCE".equals(this.jobConfig.getJobShardingStrategyType())) {
             jobFacade = new SingleShardingJobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
         } else {
             jobFacade = new ShardingJobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
         }
-
+        
         validateJobProperties();
         jobExecutor = new ElasticJobExecutor(elasticJobType, this.jobConfig, jobFacade);
         setGuaranteeServiceForElasticJobListeners(regCenter, jobListeners);
--- a/kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
+++ b/kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
@@ -83,7 +83,6 @@ public final class JobScheduler {
         setUpFacade = new SetUpFacade(regCenter, this.jobConfig.getJobName(), jobListeners);
         schedulerFacade = new SchedulerFacade(regCenter, this.jobConfig.getJobName());
         
-        // the single sharding scenario and specified SINGLE_SHARDING_BALANCE strategy
         if (1 == this.jobConfig.getShardingTotalCount()
                 && "SINGLE_SHARDING_BALANCE".equals(this.jobConfig.getJobShardingStrategyType())) {
             jobFacade = new SingleShardingJobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
@@ -105,7 +104,6 @@ public final class JobScheduler {
         setUpFacade = new SetUpFacade(regCenter, this.jobConfig.getJobName(), jobListeners);
         schedulerFacade = new SchedulerFacade(regCenter, this.jobConfig.getJobName());
         
-        // the single sharding scenario and specified SINGLE_SHARDING_BALANCE strategy
         if (1 == this.jobConfig.getShardingTotalCount()
                 && "SINGLE_SHARDING_BALANCE".equals(this.jobConfig.getJobShardingStrategyType())) {
             jobFacade = new SingleShardingJobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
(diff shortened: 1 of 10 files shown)


linghengqian (changes requested):
- The PR did not pass the code formatting validation. You can execute first `./mvnw spotless:apply -Pcheck -T1C`. Then execute `./mvnw checkstyle:check -Pcheck -T1C` to manually adjust to repair the CI.

linghengqian (changes requested):
Any updates?

baiyangzhuhong:
> Any updates?

Thanks for your advice. I follow the points what you suggested, now all code formatting validation are passed as running:
```shell
./mvnw spotless:apply -Pcheck -T1C
./mvnw checkstyle:check -Pcheck -T1C
```
The new codes are pushed, please review.

But I a little confused about that the checkstyle which caused the fields of `AbstractJobFacade` are all changed to 'private' access ability, otherwise it would fail on checkstyle testing. This is betrayed of the original intent of `AbstractJobFacade` which is a parent abstract class. Seems like it doesn't make sense and a little weired.

image

linghengqian (commented):
> But I a little confused about that the checkstyle which caused the fields of AbstractJobFacade are all changed to 'private' access ability, otherwise it would fail on checkstyle testing. This is betrayed of the original intent of AbstractJobFacade which is a parent abstract class. Seems like it doesn't make sense and a little weired.

- Do you mean you want to use private properties in the interface class? Can you explain why you want to do that? Interfaces should really only have public properties.

- CI is a bit shaky and I'm helping with retries.

linghengqian (commented):
How do I get the error in your picture? I didn't see any relevant error log in checkstyle CI.

baiyangzhuhong:
> > But I a little confused about that the checkstyle which caused the fields of AbstractJobFacade are all changed to 'private' access ability, otherwise it would fail on checkstyle testing. This is betrayed of the original intent of AbstractJobFacade which is a parent abstract class. Seems like it doesn't make sense and a little weired.
> 
> * Do you mean you want to use private properties in the interface class? Can you explain why you want to do that? Interfaces should really only have public properties.

No. I am not meaning use private properties in the Interface. 
What I am saying is there was checkstyle problems in the `AbstractJobFacade` class using 'protected' properties in my first commit codes, you know, this class is abstract class. I want the child class of `AbstractJobFacade` to extend the 'protected' properties. However, I have changed the codes to fix the checkstyle problems.

Never mind about that now, the current codes are still works fine.

> * CI is a bit shaky and I'm helping with retries.

baiyangzhuhong:
> How do I get the error in your picture? I didn't see any relevant error log in checkstyle CI.

Never mind. The error in the picture is gone cause I had changed the codes.
Thanks.

baiyangzhuhong:
> > > But I a little confused about that the checkstyle which caused the fields of AbstractJobFacade are all changed to 'private' access ability, otherwise it would fail on checkstyle testing. This is betrayed of the original intent of AbstractJobFacade which is a parent abstract class. Seems like it doesn't make sense and a little weired.
> > 
> > 
> > 
> > * Do you mean you want to use private properties in the interface class? Can you explain why you want to do that? Interfaces should really only have public properties.
> 
> No. I am not meaning use private properties in the Interface. What I am saying is there was checkstyle problems in the `AbstractJobFacade` class using 'protected' properties in my first commit codes, you know, this class is abstract class. I want the child class of `AbstractJobFacade` to extend the 'protected' properties. However, I have changed the codes to fix the checkstyle problems.
> 
> Never mind about that now, the current codes are still works fine.
> 
> > * CI is a bit shaky and I'm helping with retries.


Just for clarification,
I looked back of the checkstyle problems for first commit in the github thread, here is the original information,
image

you can reference the detail information from this address: https://github.com/apache/shardingsphere-elasticjob/pull/2462/checks?check_run_id=34090530003

thanks

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
```diff
@@ -80,7 +82,15 @@ public JobScheduler(final CoordinatorRegistryCenter regCenter, final ElasticJob
         Collection<ElasticJobListener> jobListeners = getElasticJobListeners(this.jobConfig);
         setUpFacade = new SetUpFacade(regCenter, this.jobConfig.getJobName(), jobListeners);
         schedulerFacade = new SchedulerFacade(regCenter, this.jobConfig.getJobName());
-        jobFacade = new JobFacade(regCenter, this.jobConfig.getJobName(), jobListeners, findTracingConfiguration().orElse(null));
+        
+        // the single sharding scenario and specified SINGLE_SHARDING_BALANCE strategy
```
linghengqian: Please move the comment to the method signature, or remove the comment. If you put the comment inside the method body, the compiled JavaDoc may not see it.

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
```diff
@@ -94,7 +104,15 @@ public JobScheduler(final CoordinatorRegistryCenter regCenter, final String elas
...
```
linghengqian: Please move the comment to the method signature, or remove the comment. If you put the comment inside the method body, the compiled JavaDoc may not see it.

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/sharding/strategy/type/SingleShardingBalanceJobShardingStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+ * it resolves the problem which ROUND_ROBIN is stick with the certain one job instance
+ * for the hashcode of job name is a constant value. while with SINGLE_SHARDING_BALANCE, it allows
+ * the job running on all the job instances each one by one, just like loop the job instances.
+ *
+ * this is the real round robin balance job running in the job instance dimension.
+ * </pre>
+ *
+ * @author hongzhu
+ * @version V1.0
+ * @since 2024-12-03 19:19
```
linghengqian: Please remove this personal information unless you are referencing third-party code that you do not own. If you are referencing third-party code, please make sure that the third-party has consented to your contribution of that code to the ASF.

linghengqian (commented):
- From my point of view, it doesn't make sense to use the `protected` attribute, it's better to use `private` directly. I don't think it's reasonable for the downstream to use the same java package as elasticjob.

- The current PR looks good from my perspective, but has some formatting issues.

- If you have anything to add, please provide your opinion.

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/sharding/strategy/type/SingleShardingBalanceJobShardingStrategy.java
baiyangzhuhong: That's ok.

All are done, new codes are pushed. 
The comments are not necessary, been removed.

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
baiyangzhuhong: done

kernel/src/main/java/org/apache/shardingsphere/elasticjob/kernel/internal/schedule/JobScheduler.java
baiyangzhuhong: done

linghengqian (approved):
LGTM!

PR #2491, opened 2025-12-25
Upgrade rversion

snac21:


pom.xml
```diff
@@ -552,6 +560,33 @@
             </plugins>
         </pluginManagement>
         <plugins>
+            <!-- Maven CI-Friendly Version Management -->
+            <plugin>
```
linghengqian, line 564: - Where did this CI-friendly view come from? You haven't even submitted a new issue.

linghengqian (changes requested):

pom.xml
snac21: Thank you for the label. I'd like to provide more context regarding this PR:

Consistency in Version Management: I noticed that the examples module ([link](https://github.com/apache/shardingsphere-elasticjob/blob/master/examples/pom.xml)) already uses ${revision} for version management. However, the main ElasticJob modules have not yet adopted this approach. This PR aims to unify the versioning strategy across the entire project for easier maintenance.

Fixing Build Issues: I found that elasticjob-ui currently references version 3.1.0. However, this version appears to be missing from the Maven Central Repository (possibly due to a missing release or build step in the past). By introducing ${revision} management, it will be much easier to handle project-wide packaging and ensure that all modules (including UI) are built and synchronized correctly during the CI/CD process.

linghengqian: TLDR, I personally won't agree to merge this PR. First, the next version of elasticjob is 3.1.0, so the master branch is naturally 3.1.0-SNAPSHOT, which is in line with the ASF Infra convention. If you need to compile elasticjob-ui, you have to compile elasticjob first, which will automatically publish the artifacts to the local Maven repository, and elasticjob-ui will extract the JAR from the local repository. Second, shardingsphere itself doesn't use flatten-maven-plugin, and it's strange that its sub-projects need to do so. There are dedicated Maven plugins that can update version numbers in batches. Third, you haven't applied the checkstyle rules of elasticjob for formatting.

linghengqian: If you think we need a governance framework similar to SkyWalking and SkyWalking UI that can achieve E2E across Git, the current PR doesn't reflect such a design.

snac21: Thank you for your detailed feedback. I sincerely appreciate your insights into the SkyWalking governance model and the ASF Infra conventions.

I admit that I had an insufficient understanding of these Apache project standards. Your explanation has been very helpful for my learning.

In fact, I have successfully compiled both the UI and backend locally and optimized some of the ElasticJob code. I will close this PR for now to keep the history clean.

I'll fix all Checkstyle and Spotless formatting violations.
Re-submit the optimizations through a new Issue and PR that strictly follow the project's versioning and contribution guidelines.

Thanks again for your time and guidance!

PR #2532, opened 2026-07-29
Use direct reads for lifecycle sharding assignments

### LINKED ISSUE #2531: Lifecycle statistics may return stale sharding assignments from CuratorCache
### Expected behavior

Lifecycle statistics should use the sharding assignment currently stored in the registry center.

### Actual behavior

When a running job has registered a `CuratorCache`, lifecycle statistics can return a previous
`/<jobName>/sharding/<item>/instance` value after ZooKeeper already contains the new assignment.

This can cause:

- `JobStatisticsAPI.getJobBriefInfo()` to report `SHARDING_FLAG` for a valid assignment;
- `ShardingStatisticsAPI.getShardingInfo()` to return the previous instance ID and server IP.

### Reproduction

A deterministic harness captures a real cached assignment, freezes only that path through a test adapter,
updates the same ZooKeeper path, and then invokes the real lifecycle implementations.

Observed on `0694c8f6b376b166324992be5431f6d4be131c66`:

```text
JOB_STATISTICS_STALE_ASSIGNMENT cached=10.0.0.1@-@old direct=10.0.0.2@-@new lifecycleStatus=SHARDING_FLAG
SHARDING_STATISTICS_STALE_ASSIGNMENT cached=10.0.0.1@-@old direct=10.0.0.2@-@new instanceId=10.0.0.1@-@old serverIp=10.0.0.1
RESULT jobStatisticsStaleAssignment=true shardingStatisticsStaleAssignment=true
```

The underlying real Curator cache propagation window was independently reproduced in #2526. The frozen
adapter makes the lifecycle consumer test deterministic; it does not estimate the natural window frequency.

### Environment

- ElasticJob: `3.0.6-SNAPSHOT`
- Commit: `0694c8f6b376b166324992be5431f6d4be131c66`
- Curator: `5.9.0`
- ZooKeeper client: `3.9.5`
- ZooKeeper server: `3.9.3`
- JDK: `21.0.10`

### Root cause

Both of the following methods read assignment values with cached `CoordinatorRegistryCenter.get()`:

- `JobStatisticsAPIImpl.isHasShardingFlag()`;
- `ShardingStatisticsAPIImpl.getShardingInfo()`.

ZooKeeper updates and Curator cache propagation are asynchronous, so these management queries can expose a
previous assignment.

### Proposed change

Use `CoordinatorRegistryCenter.getDirectly()` for sharding assignment values in the two lifecycle statistics
paths above. Keep `ZookeeperRegistryCenter.get()`, per-job caches, provider SPI signatures, and kernel cached
reads unchanged.

Add focused tests that preserve a stale assignment snapshot while asserting that lifecycle results follow the
authoritative assignment.

### Scope boundary

This issue only removes the local-cache stale source for assignment values. It does not promise an atomic
snapshot across assignment, instance metadata, or sharding status marker nodes. Instance deletion races should
be handled separately.

sandynz:
What: Read sharding assignment values directly in lifecycle statistics and cover stale cached values in focused tests.

Why: Lifecycle management queries should follow authoritative assignments after resharding instead of an asynchronously updated local cache.

Fixes #2531.

- Use direct reads for lifecycle sharding assignments


sandynz:
### Summary

  **Review Result: Mergeable**

  **Reason:** The latest commit fixes the cache-staleness source at both lifecycle assignment consumers without changing registry-center, kernel, API, or SPI contracts. This result is code-scope only; CI was not reviewed
  by request.

  ### Evidence

  - `JobStatisticsAPIImpl#isHasShardingFlag()` now reads `/<job>/sharding/<item>/instance` through `getDirectly()`, so job status follows the authoritative assignment rather than a potentially stale provider cache.
  - `ShardingStatisticsAPIImpl#getShardingInfo()` applies the same direct-read policy before resolving instance ID and server IP.
  - The change matches the provider contract:
    - `ZookeeperRegistryCenter#get()` may return a `CuratorCache` value.
    - `ZookeeperRegistryCenter#getDirectly()` reads through the Curator client.
    - `EtcdRegistryCenter#getDirectly()` likewise bypasses its local cache.
  - Regression tests distinguish cached and authoritative assignment values:
    - `JobStatisticsAPIImplTest` verifies that a stale cached assignment does not incorrectly produce `SHARDING_FLAG`.
    - `ShardingStatisticsAPIImplTest` verifies that the current instance ID and server IP are returned instead of the cached previous assignment.
  - No public API, SPI, configuration, dependency, packaging, or persisted-data format was changed.
  - The existing non-atomic relationship between assignment, instance metadata, and marker nodes remains outside this narrowly defined fix and is not newly introduced.

  ### Review Details

  - **Review Focus:** Code Correctness Review — CI not reviewed by request.
  - **Reviewed Scope:** `lifecycle` production and test changes in commit `7aa0a05dbdfadaafa79c3352b09b57e20143b966`; local merge-base `8ad78c68410e96f806ef4b1aa56641d245fd8f9e`. All four changed files were reviewed. A
  GitHub `/pulls/{number}/files` comparison was not applicable because the target was a local branch and commit rather than a numbered PR.
  - **Not Reviewed Scope:** GitHub Actions, live ZooKeeper/etcd integration execution, and the separate instance-traversal/deletion-race follow-up.
  - **Verification:**
    - `./mvnw -pl lifecycle -DskipITs -DskipTests=false test` — exit 0; 48 tests passed.
    - `./mvnw spotless:check -Pcheck -T1C` — exit 0.
    - `./mvnw checkstyle:check -Pcheck -T1C` — exit 0; 0 violations.
  - **Release Note / User Docs:** Not required. This is an internal lifecycle consistency correction with no user-facing configuration, API, migration, or compatibility change.

PR #2536, opened 2026-07-29
Fix etcd cache consistency and watch lifecycle management

### LINKED ISSUE #2535: EtcdRegistryCenter cache does not reflect updated or deleted keys
### Expected behavior

After a cached etcd key is updated or deleted, `EtcdRegistryCenter.get()` should eventually return the current
value or `null`.

### Actual behavior

`addCacheData()` copies the current prefix into a local map, but later writes, deletes, and watch callbacks do
not update that map. A cached key can therefore remain stale indefinitely.

Observed on `0694c8f6b376b166324992be5431f6d4be131c66`:

```text
AFTER_UPDATE cached=value-0 direct=value-1
AFTER_REMOVE cached=value-0 direct=null existed=false
RESULT updateCacheStale=true removeCacheGhost=true
```

### Reproduction

```java
center.persist("/job/config", "value-0");
center.addCacheData("/job");
center.update("/job/config", "value-1");

assertEquals("value-0", center.get("/job/config"));
assertEquals("value-1", center.getDirectly("/job/config"));

center.remove("/job/config");
assertEquals("value-0", center.get("/job/config"));
assertNull(center.getDirectly("/job/config"));
assertFalse(center.isExisted("/job/config"));
```

### Environment

- ElasticJob: `3.0.6-SNAPSHOT`
- Commit: `0694c8f6b376b166324992be5431f6d4be131c66`
- etcd server: `3.5.21`
- JDK: `21.0.10`

### Root cause

`EtcdRegistryCenter.addCacheData()` populates `Map<String, ByteSequence> cache`. `get()` returns a map entry
without contacting etcd. However, `persist()`, `update()`, `remove()`, and `EtcdWatchListener` never mutate the
map. It is only cleared by `evictCacheData()` or `close()`.

Unlike a normal asynchronous propagation delay, the stale entry has no convergence path.

### Proposed change

Define and implement a coherent etcd cache lifecycle for put, update, delete, watch, evict, and reconnect.
Add provider-level tests proving that cached reads converge after updates and stop returning deleted keys.

Please also verify transaction operations because they delegate to the same write and delete methods.

### Scope boundary

This issue concerns the etcd provider cache itself. Lifecycle APIs may independently choose authoritative reads,
but caller-side direct reads should not be the only workaround for a provider cache that never converges.

sandynz:
Start each cache watch from the revision after its initial snapshot so concurrent updates are replayed without a gap. Apply put and delete events to the cache, clear cached entries on watch failures, and close cache watchers during eviction and shutdown.

Guard cache listener shutdown against late callbacks so evicted data cannot be repopulated by an in-flight event.

Fixes #2535.

  - Keep cached etcd prefix data synchronized with PUT and DELETE events through a revision-safe snapshot/watch handoff.
  - Prevent superseded or delayed watcher callbacks from evicting refreshed cache data during watcher replacement.
  - Distinguish recoverable watch errors from terminal completion, safely handle eviction and shutdown, and avoid callback-related deadlocks when closing watchers.
  - Use striped locks to keep cache synchronization state bounded without weakening per-prefix consistency.
  - Align jetcd, gRPC, and Netty dependencies for runtime compatibility.
  - Add regression tests covering cache updates, deletions, reconnects, watcher replacement, late callbacks, shutdown, deadlock prevention, and bounded lock growth.


sandynz:
### Summary

  **Review Result: Mergeable**

  **Reason:** No code-level blockers were found within the reviewed scope. The latest two commits correctly prevent a stale watcher from evicting refreshed cache entries and bound cache lifecycle lock growth. This
  verdict reflects code-scope readiness only; CI was not reviewed by request.

  ### Evidence

  - The reviewed range is `d949032c7bf6c26e3b3b7e1285d744d2bd63cba2...dee9d2b58450954195b7a9bf6b9d47a18314856c`, containing 7 commits and 3 changed files. The worktree was clean.

  - Commit `80b123bde` serializes watcher completion and cache re-registration for the same prefix using the cache lifecycle lock. It also uses `cacheWatches.remove(prefix, this)` to ensure that an obsolete watcher
  cannot evict cache entries installed by its replacement.
  - Commit `dee9d2b58` replaces the lock map, which grew with historical cache paths, with 64 fixed lock stripes. `Math.floorMod(hashCode, 64)` always returns a value in the range `[0, 63]`, so it cannot produce a
  negative array index.
  - The implementation and focused regression tests cover snapshot revision handling, starting the watch from `revision + 1`, PUT and DELETE events, late events, watch completion, close, eviction, recoverable errors, and
  watcher replacement.
  - In jetcd 0.7.7, retryable watch errors resume from the recorded revision, while non-retryable errors close the watcher and trigger completion. Therefore, retaining the cache in `onError` and invalidating it in
  `onCompleted` is consistent with jetcd's lifecycle.
  - The resolved dependency tree uses gRPC `1.82.2` and consistently resolves Netty modules to `4.1.136.Final`. The official gRPC compatibility table places gRPC `1.81.x` and later on the Netty 4.1 line, and Netty
  `4.1.136.Final` is an official release. Together with the resolved dependency tree and consumer tests, no version conflict was found. [gRPC Netty compatibility](https://github.com/grpc/grpc-java/blob/v1.82.2/
  SECURITY.md#netty), [Netty 4.1.136.Final release](https://github.com/netty/netty/releases/tag/netty-4.1.136.Final)
  - All checks passed.
  - No significant issues were found in completeness, correctness, or coherence.
  - The final adversarial pass found no additional independent actionable findings.

  ### Review Details

  - **Review Focus:** Code Correctness Review — CI not reviewed by request.
  - **Reviewed Scope:** `pom.xml`, the etcd provider's `EtcdRegistryCenter`, and its new test class. The reviewed HEAD was `dee9d2b58450954195b7a9bf6b9d47a18314856c`, with merge base
  `d949032c7bf6c26e3b3b7e1285d744d2bd63cba2`. Particular attention was given to commits `80b123bde` and `dee9d2b58`.
  - **GitHub File Match:** No PR number was provided for this review, so the GitHub `/pulls/{number}/files` list was not available for comparison. The local triple-dot file list matched the user-specified range.
  - **Not Reviewed Scope:** GitHub Actions, the cross-platform and cross-JDK CI matrix, and a real-etcd integration test were not reviewed in this round. The affected workflows are `Required - Check` and `Java CI with
  Maven`; the GraalVM workflow path filters do not match these changes.
  - **Verification:**
    - IDE `EtcdRegistryCenterTest`: exit `0`; 11 tests passed.
    - IDE targeted build of the changed production and test classes: successful.
    - `./mvnw -pl registry-center/provider/etcd -DskipTests spotless:check -Pcheck -T1C`: exit `0`.
    - `./mvnw -pl registry-center/provider/etcd -DskipTests checkstyle:check -Pcheck -T1C`: exit `0`.
    - `./mvnw -pl registry-center/provider/etcd -DskipTests apache-rat:check -Pcheck -T1C`: exit `0`.
    - etcd provider dependency tree: exit `0`; jetcd `0.7.7`, gRPC `1.82.2`, and Netty `4.1.136.Final`.
    - `./mvnw -pl restful,spring --batch-mode --no-transfer-progress test -Pcheck -T1C`: exit `0`; `BUILD SUCCESS`.
    - The first dependency-tree invocation exited before Maven started because an unquoted wildcard was expanded by zsh. The single corrected rerun with the argument quoted passed; this was not a code failure.
  - **Release Note / User Docs:** Not required. The changes affect internal provider cache consistency, resource lifecycle, and dependency alignment without changing public APIs, SPIs, configuration, or migration
  contracts.

sandynz:
## Additional findings from implementation and validation

The original reproduction proved that the etcd cache was only a one-time snapshot and therefore had no convergence
path after a key was updated or deleted. Implementing and validating the fix exposed several additional lifecycle
requirements. Updating the local map from watch events is necessary, but it is not sufficient by itself.

### The snapshot/watch handoff must be revision-safe

Cache initialization consists of two operations: reading the current prefix and starting a prefix watch. If the
watch starts without a revision derived from the snapshot, an update committed between those operations can be
missed permanently.

The cache lifecycle therefore needs to:

1. read the prefix snapshot and its etcd revision `R`;
2. populate the local cache from that snapshot; and
3. start the prefix watch at revision `R + 1`.

This makes the snapshot and the subsequent watch a continuous logical stream. PUT and DELETE events from that watch
then update the cached values and remove deleted keys.

### Replacing a watcher introduces another stale-cache race

Re-registering the same cache prefix closes the previous watcher and installs a new snapshot/watch pair. A previous
watcher's completion callback can nevertheless be delayed until after the replacement has populated fresh data.
Without ownership checks, that stale callback can evict the replacement cache and remove the current watcher
registration.

The completion path must therefore invalidate a cache only if its listener is still the currently registered
listener for that prefix. Registration, replacement, completion, and explicit eviction for the same prefix must also
share a stable synchronization boundary. This prevents a superseded watcher from deleting data owned by its
replacement.

### Recoverable watch errors and terminal completion have different semantics

In jetcd 0.7.7, a recoverable watch failure invokes `Watch.Listener.onError()` and the client can reschedule the watch
from its recorded revision. A terminal failure closes the watcher and eventually invokes `onCompleted()`.
Consequently:

- `onError()` must retain the existing snapshot while jetcd reconnects;
- `onCompleted()` must invalidate the listener-owned cache and restore direct-read behavior;
- explicit eviction and registry-center shutdown must deactivate the listener before closing its jetcd watcher, so
  late callbacks cannot repopulate an evicted cache; and
- `Watcher.close()` must not run while holding the listener monitor, because jetcd may synchronously call back into
  the listener during close and otherwise create a lock-order deadlock.

This behavior was verified against the published jetcd 0.7.7 source, specifically `WatchImpl.handleError()` and
`WatchImpl.WatcherImpl.close()`.

### Synchronization state must remain bounded

A stable per-prefix lock prevents the replacement race, but retaining one lock object for every historical cache path
causes memory usage to grow with every job ever observed, even after its cache and watcher are evicted.

The implementation therefore uses a fixed set of striped locks. This keeps synchronization state bounded while
preserving serialization for the same prefix. Hash collisions may serialize unrelated prefixes, but they affect
contention only; cache correctness is unchanged. Active watchers still scale with active cached prefixes and are
closed by explicit eviction or registry-center shutdown.

### Runtime validation also exposed a separate dependency-alignment issue

The cache defect itself is independent of dependency versions. However, exercising the provider with the current
jetcd client also exposed an incompatible mixture of the Netty 4.2 line and gRPC/jetcd dependencies built on the
Netty 4.1 line. The dependency set was aligned to:

- jetcd `0.7.7`;
- gRPC `1.82.2`; and
- Netty `4.1.136.Final`.

The gRPC Java security documentation recommends keeping all unshaded Netty artifacts on one compatible version and
lists Netty `4.1.132.Final` as the baseline for gRPC `1.81.x` and later. Netty `4.1.136.Final` is a later release on
that same 4.1 line. See the
[`grpc-java` Netty compatibility table](https://github.com/grpc/grpc-java/blob/v1.82.2/SECURITY.md#netty) and the
[`netty-4.1.136.Final` release](https://github.com/netty/netty/releases/tag/netty-4.1.136.Final).

This alignment is a runtime compatibility prerequisite discovered during validation, not the root cause of the
stale cache.

### Regression coverage added for the complete lifecycle

The resulting provider-level tests cover:

- starting the watch at the snapshot revision plus one;
- applying watched PUT and DELETE events;
- falling back to direct reads after terminal watch completion;
- retaining cached data across a recoverable watch reconnect;
- preventing a previous watcher's delayed completion from evicting refreshed data;
- ignoring late events after explicit eviction;
- closing cache watchers during registry-center shutdown;
- avoiding deadlock when watcher close overlaps an error callback; and
- keeping cache synchronization locks bounded after many distinct cache paths.

### Updated conclusion

The complete fix is not only "update the map from watch callbacks." It establishes a cache lifecycle protocol with a
revision-safe snapshot/watch handoff, current-listener ownership, distinct reconnect and terminal-completion
behavior, callback-safe shutdown, and bounded synchronization state. These constraints prevent both the original
indefinite stale read and the less obvious races introduced by watcher replacement and cleanup.

PR #2537, opened 2026-09-03
Fix excldue typo in apache-rat exclude patterns

tanvir-ux:
## What
Fixed `<excldue>` → `<exclude>` in the apache-rat plugin config.

Maven silently ignores unknown tags, so `**/*.MD` (and any other patterns under the misspelled tag) were never excluded from license checks.

## Why
Same class of bug as apache/shenyu#7028 / related rat typo reports. Small, safe pom-only fix.

## Test
Diff-only change to POM exclude tags; no runtime behavior change.


linghengqian (approved):
LGTM.

